PR #1, opened 2016-11-02
Configuration parse fix

mjarco:


medzin:
@mjarco Any chance for some CI build for this?

config/config.go
```diff
@@ -102,21 +120,16 @@ func Configure() (conf Config, err error) {
...
-		}
-		conf.BackendURLs = append(conf.BackendURLs, url)
-	}
-
-	if len(conf.MaintainedBackend) > 0 {
-		u, perr := url.Parse(conf.MaintainedBackend)
-		if perr != nil {
-			return conf, perr
+	confFile, openErr := os.Open(*confFilePath)
+	if openErr == nil {
```
medzin: Invert `if` statement to reduce nesting.

httphandler/httphandler.go
```diff
@@ -89,10 +90,13 @@ func NewHandler(conf config.Config) http.Handler {
 		Dial:                dialer.Dial,
 		DisableKeepAlives:   conf.KeepAlive,
 		MaxIdleConnsPerHost: int(conf.ConnLimit)}
-
+	backends := make([]*url.URL, len(conf.Backends))
+	for _, backend := range(conf.Backends) {
+		backends = append(backends, backend.URL)
```
medzin: Why appending? You already have a slice with required length, so you can just assign values to indexes.

config/config.go
```diff
@@ -36,19 +36,40 @@ type YamlConfig struct {
...
+
+
+func (j *YAMLURL) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var s string
+	if err := unmarshal(&s); err != nil {
+		return err
+	}
+	url, err := url.Parse(s)
+	if url.Host == "" {
+		return fmt.Errorf("url should match proto://host[:port]/path scheme, got %q", s)
```
medzin: How this message will look when the `s` variable is empty?

config/config.go
```diff
@@ -102,21 +120,16 @@ func Configure() (conf Config, err error) {
...
+	confFile, openErr := os.Open(*confFilePath)
+	if openErr == nil {
+		yconf, parseErr := parseConf(confFile)
+		if parseErr != nil {
+			return conf, parseErr
 		}
-		conf.MaintainedBackend = u.Host
+		conf = Config{YamlConfig: yconf}
+	} else {
+		fmt.Println("Cannot read config file:", openErr.Error())
```
mjarco: I'll remove above print. I would rather use log.Fatal, but as all loggers log to syslog error message would not be visible to user. So I'll leave fmt.Print in main.go and return in main func

config/config.go
mjarco: It will never run this function in that case.

medzin: Can you write test for case when Host will be empty?

config/config_test.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"github.com/go-yaml/yaml"
+	"testing"
+)
+
+type TestYaml struct {
+	Field YAMLURL
+}
+
+func TestYAMLURLParsing(t *testing.T) {
```
medzin: Split to two functions with more meaningful test names. Something like `TestIfParsesCorrectYAML`, `TestIfParsesIncorrectYAML`.

config/config_test.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		t.Error(err.Error())
+	}
+}
+
+func TestYAMLURLParsingFailure(t *testing.T) {
+	incorrect := []byte(`field: golang.org:80/pkg/net`)
+	testyaml := TestYaml{}
+	err := yaml.Unmarshal(incorrect, &testyaml)
+	if err == nil {
```
medzin: `assert.Error(err)`

config/config_test.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func TestYAMLURLParsingEmpty(t *testing.T) {
+	incorrect := []byte(`field:`)
+	testyaml := TestYaml{}
+	err := yaml.Unmarshal(incorrect, &testyaml)
+	if err != nil {
+		t.Errorf("Should not even try to parse")
+	}
+	if testyaml.Field.URL != nil {
```
medzin: `assert.Nil()`

config/config.go
```diff
@@ -36,19 +36,37 @@ type YamlConfig struct {
 	//List request methods to be logged in synclog in case of backend failure
 	SyncLogMethods []string `yaml:"SyncLogMethods,omitempty"`
 	//Should we keep alive connections with backend servers
-	KeepAlive bool `yaml:"SyncLogMethods,omitempty"`
+	KeepAlive bool `yaml:"KeepAlive,omitempty"`
```
wendigo: I guess that omitempty does not make sense here - zero value of bool is false

config/config_test.go
```diff
@@ -14,28 +15,20 @@ func TestYAMLURLParsingSuccessful(t *testing.T) {
...
 	incorrect := []byte(`field:`)
 	testyaml := TestYaml{}
 	err := yaml.Unmarshal(incorrect, &testyaml)
-	if err != nil {
-		t.Errorf("Should not even try to parse")
-	}
-	if testyaml.Field.URL != nil {
-		t.Errorf("Should be nil")
-	}
+	assert.Nil(t, err, "Should not even try to parse")
```
medzin: `assert.NoError`

config/config_test.go
```diff
@@ -14,28 +15,20 @@ func TestYAMLURLParsingSuccessful(t *testing.T) {
 	correct := []byte(`field: http://golang.org:80/pkg/net`)
 	testyaml := TestYaml{}
 	err := yaml.Unmarshal(correct, &testyaml)
-	if err != nil {
-		t.Error(err.Error())
-	}
+	assert.Nil(t, err, "Should be correct")
```
medzin: `assert.Error`

PR #8, opened 2016-11-18
Json logs

mjarco:
AccessLog has json entries instead flat csv line
Fixed issue with response headers


httphandler/roundtripper_decorators.go
```diff
@@ -36,8 +36,11 @@ func (lrt *loggingRoundTripper) RoundTrip(req *http.Request) (resp *http.Respons
 		statusCode,
 		duration,
 		errStr)
-
-	lrt.accessLog.Println(accessLogMessage.String())
+	jsonb, almerr := accessLogMessage.JSON()
+	if almerr != nil {
+		log.Println(almerr.Error())
+	}
+	lrt.accessLog.Printf("%s", jsonb)
```
medzin: It will be always executed, even if `JSON()` method fails - so it will log an empty string?

httphandler/roundtripper_decorators_test.go
```diff
@@ -100,7 +100,9 @@ func TestAccessLogging(t *testing.T) {
 	sendReq(t, srv, "GET", nil, rt)
 
 	amdstr := strings.Trim(buf.String(), "\n")
```
medzin: You can use `Bytes()` and `bytes.Trim()` functions here, so that subsequent conversion to byte array will not be needed.

httphandler/log.go
```diff
@@ -48,6 +51,44 @@ func ScanCSVAccessLogMessage(csvstr string) (AccessMessageData, error) {
...
+		slmd.Method,
+		slmd.FailedHost,
+		slmd.Path,
+		slmd.SuccessHost,
+		slmd.UserAgent,
+		slmd.ErrorMsg)
+}
+
+//JSON produces data in json format with fields in following order
+func (slmd SyncLogMessageData) JSON() ([]byte, error) {
```
medzin: Maybe you want to implement an [json.Marshaler](https://golang.org/pkg/encoding/json/#Marshaler) interface here?

httphandler/roundtripper_decorators.go
```diff
@@ -36,8 +36,11 @@ func (lrt *loggingRoundTripper) RoundTrip(req *http.Request) (resp *http.Respons
 		statusCode,
 		duration,
 		errStr)
-
-	lrt.accessLog.Println(accessLogMessage.String())
+	jsonb, almerr := accessLogMessage.JSON()
+	if almerr != nil {
+		log.Println(almerr.Error())
```
medzin: Add more meaningful log message here.

PR #12, opened 2016-12-15
fixing no error when port already in use

bieli:


main.go
```diff
@@ -67,7 +67,7 @@ func (s *service) start() error {
 	listener, err := net.Listen("tcp", s.config.Listen)
 
 	if err != nil {
-		return err
+		panic(err)
```
medzin: Change to `Fatal` here [main.go:48] (https://github.com/allegro/akubra/blob/master/main.go#L48) instead.

mjarco: It's still pointless. If server doesn't start main will finish, so Fatal does not change anything.

Real issue here is that all log messages are passed to syslog instead stdout. Best solution would be to add syslog/stdout switch in command line arguments an changes in config/config.go.

bieli: My simple change it's important for external integration tool.
Big change with logrus module need more work with refactoring.
Today akubra `return` in this case give me `exitcode 0` but when we add `panic - exitcode is 2` and it's temporary solution for me :-)

medzin: @bieli You should panic then in `main()` function in response to error returned from `start()` function. More precisely - you should use  [Fatalf()](https://golang.org/pkg/log/#Fatalf) there instead of [Printf](https://golang.org/pkg/log/#Printf).

PR #24, opened 2017-02-03
Wait for all backends if performing bucket operation

mjarco:


httphandler/response_merger.go
```diff
@@ -91,19 +82,34 @@ func (rd *responseMerger) _handle(in <-chan *transport.ReqResErrTuple, out chan<
...
-	respPassed := false
+	firstPassed := false
 
 	for {
 		r, hasMore := <-in
 		if !hasMore {
 			break
 		}
+
+		errorMsg := "No error"
```
janisz: If no error then this should be empty.

httphandler/response_merger.go
```diff
@@ -91,19 +82,34 @@ func (rd *responseMerger) _handle(in <-chan *transport.ReqResErrTuple, out chan<
...
+		if r.Err != nil {
+			errorMsg = r.Err.Error()
+		}
+
+		rd.runtimeLog.Printf("RGW resp %q, %q, %q, %t, %q",
```
janisz: What's `RGW`?

httphandler/response_merger.go
```diff
@@ -91,19 +82,34 @@ func (rd *responseMerger) _handle(in <-chan *transport.ReqResErrTuple, out chan<
...
+			r.Req.URL.Path,
+			r.Req.Method,
+			r.Req.Host,
+			r.Failed,
+			errorMsg)
+
 		// pass first successful answer to client
```
janisz: This comment is not true

PR #26, opened 2017-02-16
Maxbodysize

ksitak:
prevent from exceeding max body size

### DIFF
--- a/httphandler/httphandler.go
+++ b/httphandler/httphandler.go
@@ -7,9 +7,12 @@ import (
 	"io"
 	"net/http"
 	"time"
+	"strconv"
+	"errors"
 
 	"github.com/allegro/akubra/config"
 	"github.com/allegro/akubra/log"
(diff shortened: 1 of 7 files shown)


httphandler/httphandler.go
```diff
@@ -62,6 +73,22 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}()
 }
 
+func (h *Handler) validateIncomingRequest(req *http.Request) int {
+	var contentLength int64
+	contentLengthHeader := req.Header.Get("Content-Length")
+	if contentLengthHeader != "" {
+		var err error
+		contentLength, err = strconv.ParseInt(contentLengthHeader, 10, 64)
```
mjarco: replace above 2 lines with
contentLength, err := strconv.ParseInt(contentLengthHeader, 10, 64)

PR #27, opened 2017-02-16
Content length to synclog

mjarco:


httphandler/response_merger.go
```diff
@@ -43,6 +43,7 @@ func (rd *responseMerger) synclog(r, successfulTup *transport.ReqResErrTuple) {
 		successfulTup.Req.URL.Path,
 		successfulTup.Req.Host,
 		r.Req.Header.Get("User-Agent"),
+		successfulTup.Res.Header.Get("Content-Length"),
```
janisz: What if header is not provided? Do we need a fallback?

httphandler/log.go
```diff
@@ -57,33 +57,35 @@ func ScanCSVAccessLogMessage(csvstr string) (AccessMessageData, error) {
...
-	UserAgent   string `json:"useragent"`
-	ErrorMsg    string `json:"error"`
-	ReqID       string `json:"reqID"`
-	Time        string `json:"ts"`
+	Method        string `json:"method"`
+	FailedHost    string `json:"failedhost"`
+	Path          string `json:"path"`
+	SuccessHost   string `json:"successhost"`
+	UserAgent     string `json:"useragent"`
+	ContentLength string `json:"content-length"`
```
janisz: Why `string` not `uint`?

httphandler/response_merger.go
mjarco: yep. fixed

httphandler/log.go
mjarco: will be int64 negative number will indicate header was missing

httphandler/log.go
```diff
@@ -57,33 +57,35 @@ func ScanCSVAccessLogMessage(csvstr string) (AccessMessageData, error) {
...
-	UserAgent   string `json:"useragent"`
-	ErrorMsg    string `json:"error"`
-	ReqID       string `json:"reqID"`
-	Time        string `json:"ts"`
+	Method        string `json:"method"`
+	FailedHost    string `json:"failedhost"`
+	Path          string `json:"path"`
+	SuccessHost   string `json:"successhost"`
+	UserAgent     string `json:"useragent"`
+	ContentLength int64  `json:"content-length"`
```
janisz: Comment about negative value meaning would be nice.

PR #28, opened 2017-02-22
Fixed: connection leaks if cluster recursion occured

mjarco:


sharding/shards_ring.go
```diff
@@ -95,6 +96,12 @@ func (sr shardsRing) regressionCall(cl cluster, req *http.Request) (string, *htt
 	if (err != nil || resp.StatusCode > 400) && req.Method != http.MethodPut {
 		rcl, ok := sr.clusterRegressionMap[cl.name]
 		if ok {
+			_, copyErr := io.Copy(ioutil.Discard, resp.Body)
```
janisz: Please add a doc why it's needed and why not calling `resp.Body.Close()`

mjarco: resp.Body.Close() works as well

PR #34, opened 2017-03-10
Replaced hostname with fqdn in metrics path

mjarco:


metrics/metrics.go
```diff
@@ -135,7 +136,13 @@ func appendDefaults(prefix string) (string, error) {
 }
 
 // stubbed out for testing
-var hostname = os.Hostname
+var hostname = func() (string, error) {
+	out, err := exec.Command("hostname", "-f").Output()
```
bieli: maybe https://github.com/ShowMax/go-fqdn/blob/master/fqdn.go is better solution ?

ksitak: No, it sucks. If my host doesn't have fqdn then akubra doesn't start (and reports much saying message:time="Mar 10 13:33:06.209397" level=fatal msg="exit status 1").
I prefer it to fallback to os.Hostname if fqdn is not available.

metrics/metrics.go
```diff
@@ -137,11 +139,7 @@ func appendDefaults(prefix string) (string, error) {
 
 // stubbed out for testing
 var hostname = func() (string, error) {
```
ksitak: do we still need to return not-possible error?

PR #37, opened 2017-04-05
Psql synclogger

mjarco:


log/sql/connection.go
```diff
@@ -0,0 +1,80 @@
...
+	db, err := NewConnection(config)
+	if err != nil {
+		return nil, err
+	}
+	return NewSyncLogDBHook(db, config)
+}
+
+func buildQuery(tmpl *template.Template, jsonStr string) (string, error) {
+	substitutionMap := make(map[string]string)
+	err := json.Unmarshal([]byte(jsonStr), &substitutionMap)
```
ksitak: fails for any non-string json fields, eg content-length

mjarco: Should be fine now

ksitak:
Can you add some metrics to count successful and failed db inserts?

log/sql/connection.go
```diff
@@ -0,0 +1,86 @@
...
+			metrics.Mark("reqs.global.inconsistencies.store-failure")
+			return errt
+		}
+		_, erre := tx.Exec(query)
+		if erre != nil {
+			metrics.Mark("reqs.global.inconsistencies.store-failure")
+			return erre
+		}
+		errc := tx.Commit()
+		if errc != nil {
```
ksitak: IMO it's not ok... we should count success when there's nil here, failure otherwise.

log/sql/connection.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		_, erre := tx.Exec(query)
+		if erre != nil {
+			metrics.Mark("reqs.global.inconsistencies.store-failure")
+			return erre
+		}
+		errc := tx.Commit()
+		if errc == nil {
+			metrics.Mark("reqs.global.inconsistencies.store-success")
+		}
```
ksitak: Why we don't increment store-failure when commit fails?

log/sql/connection.go
```diff
@@ -0,0 +1,88 @@
...
+			metrics.Mark("reqs.global.inconsistencies.store-failure")
+			return erre
+		}
+		errc := tx.Commit()
+		if errc == nil {
+			metrics.Mark("reqs.global.inconsistencies.store-success")
+		} else {
+			metrics.Mark("reqs.global.inconsistencies.store-failure")
+		}
+		return errc
```
ksitak: For me it would look much more intuitive if it was analogous to other cases above. Something like this:

`errc := tx.Commit()`
`if errc != nil {`
`	metrics.Mark("reqs.global.inconsistencies.store-failure")`
`	return errc`
`}`
`metrics.Mark("reqs.global.inconsistencies.store-success")`
`return nil`

metrics/metrics.go
```diff
@@ -80,36 +78,34 @@ func Init(cfg Config) (err error) {
...
 }
 
-func startExpvar(cfg Config, handler http.Handler) {
-	err := http.ListenAndServe(cfg.ExpAddr, handler)
-	if err != nil {
-		log.Printf("Could not start exp server: %q", err.Error())
-	}
+func startExpvar(cfg Config, handler http.Handler) error {
+	started := make(chan error)
+	go func() {
```
janisz: Why do we need this gorutine? [`ListenAndServe`](https://golang.org/pkg/net/http/#ListenAndServe) is blocking and "always returns a non-nil error".

PR #38, opened 2017-04-07
Validate config technical HTTP endpoint

bieli:
For CI proccess we need technical/validation HTTP endpoint for validation Akubra Yaml configuration.
When we POST akubra config file:
```
curl -vv -X POST -H "Content-Type: application/yaml" --data-binary @test.cfg.yaml http://127.0.0.1:8071/validate/configuration
```
W got possible responses:
- HTTP 400 and info in body about validation errors
- HTTP 200 when pass validation

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -18,6 +18,7 @@ import (
 type YamlConfig struct {
 	// Listen interface and port e.g. "0.0.0.0:8000", "127.0.0.1:9090", ":80"
 	Listen string `yaml:"Listen,omitempty" validate:"regexp=^(([0-9]+[.][0-9]+[.][0-9]+[.][0-9]+)?[:][0-9]+)$"`
+	TechnicalEndpointListen string `yaml:"TechnicalEndpointListen,omitempty" validate:"regexp=^(([0-9]+[.][0-9]+[.][0-9]+[.][0-9]+)?[:][0-9]+)$"`
 	// List of backend URI's e.g. "http://s3.mydatacenter.org"
 	Backends []shardingconfig.YAMLUrl `yaml:"Backends,omitempty,flow"`
 	// Maximum accepted body size
--- a/config/config.go
+++ b/config/config.go
@@ -3,8 +3,10 @@ package config
 import (
 	"io"
 	"io/ioutil"
+	"net/http"
 	"os"
 
+	"fmt"
 	"github.com/allegro/akubra/log"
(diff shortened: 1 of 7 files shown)


main.go
```diff
@@ -104,3 +105,18 @@ func (s *service) start() error {
 func newService(cfg config.Config) *service {
 	return &service{conf: cfg}
 }
+
+func (s *service) startTechnicalEndpoint(conf config.Config) {
+	log.Printf("Starting technical HTTP endpoint on port: %q", conf.TechnicalEndpointListen)
+	serveMuxHandler := http.NewServeMux()
+	serveMuxHandler.HandleFunc(
+		"/validate/configuration",
```
ksitak: 1. IMO it should be /configuration/validate. In the future we may want to add some other actions regarding configuration, like /configuration/update
2. Akubra never terminates connection itself on this port - it sucks.
3. There's no body size limit - we can easily kill akubra uploding too large content.
4. When I run akubra from this branch... well... it opens technical port, that's nice. But it does not open it's ordinary port to handle s3 requests... wtf?

ksitak:
Just one more remark - when I contact technical endpoint with method other than post it should end up with 405 Method Not Allowed, not 400 Bad Request.

config/config.go
```diff
@@ -14,10 +17,17 @@ import (
...
+
+// TechnicalEndpointHeaderContentType for /configuration/validate endpoint
+const TechnicalEndpointHeaderContentType = "application/yaml"
+
 // YamlConfig contains configuration fields of config file
 type YamlConfig struct {
 	// Listen interface and port e.g. "0.0.0.0:8000", "127.0.0.1:9090", ":80"
-	Listen string `yaml:"Listen,omitempty" validate:"regexp=^(([0-9]+[.][0-9]+[.][0-9]+[.][0-9]+)?[:][0-9]+)$"`
+	Listen                  string `yaml:"Listen,omitempty" validate:"regexp=^(([0-9]+[.][0-9]+[.][0-9]+[.][0-9]+)?[:][0-9]+)$"`
+	TechnicalEndpointListen string `yaml:"TechnicalEndpointListen,omitempty" validate:"regexp=^(([0-9]+[.][0-9]+[.][0-9]+[.][0-9]+)?[:][0-9]+)$"`
```
bieli: I suggest to create new PR for update `Listen and TechnicalEndpointListen` validation type/method.

PR #50, opened 2017-06-09
Regions support

ksitak:


config/config_test.go
```diff
@@ -347,9 +305,20 @@ func TestShouldNotPassValidateConfigurationHTTPHandlerWithWrongRequests(t *testi
...
+	result, _ := ValidateConf(testConfData, true)
+
+	assert.False(t, result)
+}
+
 func PrepareYamlConfig(bodyMaxSize shardingconfig.HumanSizeUnits, idleConnTimeoutInp time.Duration, responseHeaderTimeoutInp time.Duration,
-	maintainedBackendHost string, listen string, technicalEndpointListen string, clientCfgName string,
-	clientClusters []string) YamlConfig {
+	maintainedBackendHost string, listen string, technicalEndpointListen string,
+	regions map[string]shardingconfig.RegionConfig) YamlConfig {
```
janisz: Too many parameters! Move them to structure.

config/validator.go
```diff
@@ -60,25 +60,38 @@ func UniqueValuesInSliceValidator(v interface{}, param string) error {
...
 
-// ClientClustersEntryLogicalValidator validate Client->Clusters entry and make sure that all required clusters are defined
-func (c *YamlConfig) ClientClustersEntryLogicalValidator(valid *bool, validationErrors *map[string][]error) {
-	errorsList := make(map[string][]error)
-
-	if len(c.Client.Clusters) == 0 {
-		*valid = false
-		errorDetail := []error{errors.New("Empty clusters definition")}
-		errorsList["ClientClustersEntryLogicalValidator"] = errorDetail
+//RegionsEntryLogicalValidator validator for Regions part
```
janisz: This does not tell anything. Please describe what/why it's doing

config/validator.go
```diff
@@ -60,25 +60,38 @@ func UniqueValuesInSliceValidator(v interface{}, param string) error {
...
+func (c *YamlConfig) RegionsEntryLogicalValidator(valid *bool, validationErrors *map[string][]error) {
```
janisz: output parameters? This should be return of this function.

sharding/config/config.go
```diff
@@ -13,21 +13,23 @@ import (
...
+	Cluster string `yaml:"Cluster"`
+	//Cluster weight
+	Weight int `yaml:"Weight"`
+}
+
+//RegionConfig region configuration
+type RegionConfig struct {
+	//Multi cluster config
+	Clusters []MultiClusterConfig `yaml:"Clusters"`
+	//domains used for region matching
```
janisz: `d` → `D`

sharding/sharding.go
```diff
@@ -1,85 +1,28 @@
...
-	if ok {
-		return s3cluster, nil
-	}
-	s3cluster, err := rf.initCluster(name)
-	if err != nil {
-		return s3cluster, err
-	}
-	rf.clusters[name] = s3cluster
-	return s3cluster, nil
+	storages  *storages.Storages
```
janisz: why using pointer?

sharding/shards_ring.go
```diff
@@ -133,8 +139,8 @@ func (sr *ShardsRing) logInconsistency(key, expectedClusterName, actualClusterNa
 	}
 }
 
-// RoundTrip performs http requests
-func (sr ShardsRing) RoundTrip(req *http.Request) (resp *http.Response, rerr error) {
+//DoRequest performs http requests
```
janisz: What it does? What it returns and when? How it handle nil? Compare with https://golang.org/pkg/net/http/#Client.Do

storages/storages.go
```diff
@@ -0,0 +1,71 @@
...
+func (st Storages) initCluster(name string) (Cluster, error) {
+	clusterConf, ok := st.Conf.Clusters[name]
+	if !ok {
+		return Cluster{}, fmt.Errorf("no cluster %q in configuration", name)
+	}
+	respHandler := httphandler.EarliestResponseHandler(st.Conf)
+	return newMultiBackendCluster(st.Transport, respHandler, clusterConf, name, st.Conf.MaintainedBackends), nil
+}
+
+//GetCluster gets cluster by name
```
janisz: define what it returns and when error is not nil

config/config_test.go
ksitak: It's now one less than before!

config/validator.go
ksitak: Really? It validates "regions" part of configuration file... and tells if it's ok or not...

PR #52, opened 2017-08-02
Bugfix for regions validation

bieli:


config/config.go
```diff
@@ -174,8 +174,13 @@ func ValidateConf(conf YamlConfig, enableLogicalValidator bool) (bool, map[strin
 	validator.SetValidationFunc("UniqueValuesSlice", UniqueValuesInSliceValidator)
 	valid, validationErrors := validator.Validate(conf)
 	if valid && enableLogicalValidator {
-		conf.RegionsEntryLogicalValidator(&valid, &validationErrors)
-		conf.ListenPortsLogicalValidator(&valid, &validationErrors)
+		var validRegions bool
+		var validListenPorts bool
+		conf.RegionsEntryLogicalValidator(&validRegions, &validationErrors)
+		conf.ListenPortsLogicalValidator(&validListenPorts, &validationErrors)
+		if !validRegions || !validListenPorts {
```
mjarco: How about removing this negative if with
```
conf.RegionsEntryLogicalValidator(&valid, &validationErrors)
conf.ListenPortsLogicalValidator(&validListenPorts, &validationErrors)
valid = valid && validListenPorts

```
?

log/logger.go
```diff
@@ -11,10 +11,12 @@ import (
...
 )
 
-const (
-	// ContextreqIDKey is Request Context Value key for debug logging
-	ContextreqIDKey = "reqID"
-)
+// ContextKey for logging context information
+type ContextKey string
+
+func (c ContextKey) String() string {
```
mjarco: What was wrong with old construction. Why you're implementing Stringer for named string type? And then not use this metrod?

bieli: goling + `vet` alerting

config/config.go
bieli: Ok, we can removing 3 lines from code in your idea

PR #53, opened 2017-08-04
Adding health check endpoint

bieli:


httphandler/roundtripper_decorators.go
```diff
@@ -6,10 +6,17 @@ import (
...
 
+	"io/ioutil"
+
 	"github.com/allegro/akubra/log"
 	shardingconfig "github.com/allegro/akubra/sharding/config"
 )
 
+const (
+	// HealthCheckEndpoint is for load balancers and other related systems
+	HealthCheckEndpoint string = "/status/ping"
```
mjarco: It would be nice to define uri in configuration

httphandler/roundtripper_decorators.go
```diff
@@ -122,6 +129,33 @@ func (os optionsHandler) RoundTrip(req *http.Request) (resp *http.Response, err
...
 }
 
+type statusHandler struct {
+	roundTripper http.RoundTripper
+}
+
+func (sh statusHandler) RoundTrip(req *http.Request) (resp *http.Response, err error) {
+
+	if strings.Contains(req.Method, http.MethodGet) && strings.Contains(strings.ToLower(req.URL.Path), HealthCheckEndpoint) {
+		resp, err = sh.roundTripper.RoundTrip(req)
```
mjarco: We should discuss if HC should be passed further or should it be terminated in akubra

httphandler/roundtripper_decorators.go
bieli: Done

PR #54, opened 2017-08-07
Weights must be float (0.0 to 1.0)

ksitak:


sharding/config/config.go
```diff
@@ -21,7 +21,7 @@ type MultiClusterConfig struct {
 	//Cluster name
 	Cluster string `yaml:"Cluster"`
 	//Cluster weight
-	Weight int `yaml:"Weight"`
+	Weight float64 `yaml:"Weight"`
```
janisz: Looks like breaking change

mjarco: Weight float64 `yaml:"Weight" validate:"min=0,max=1"`

ksitak: We already have validator for that. And putting code above does not work "out of the box".

PR #56, opened 2017-08-30
Mark region as default

mjarco:


sharding/config/config.go
```diff
@@ -30,6 +30,7 @@ type RegionConfig struct {
 	Clusters []MultiClusterConfig `yaml:"Clusters"`
 	//Domains used for region matching
 	Domains []string `yaml:"Domains"`
+	Default bool     `yaml:"Default,omitempty"`
```
janisz: Please add godoc

mjarco: thnx. done

PR #58, opened 2017-11-07
Rewrite config validators

mjarco:


main.go
```diff
@@ -86,8 +87,11 @@ func (s *service) start() error {
 	}
 	// TODO: Decorate ^ roundtripper here now - fix accesslog in configuration
 	syncLog, err := log.NewDefaultLogger(s.config.Logging.Synclog, "LOG_LOCAL1", true)
-
-	respHandler := httphandler.LateResponseHandler(syncLog, s.config.Logging.SyncLogMethodsSet)
+	methods := make([]interface{}, 0, len(s.config.Logging.SyncLogMethods))
```
janisz: Why `[]interface{}` and not `[]string`?

bieli: probably by usage: `set.NewSetFromSlice(methods)` -> `func NewSetFromSlice(s []interface{}) Set`

PR #61, opened 2017-11-14
GitHub tag badge in Readme

platan:
This project does not have GitHub releases, so GitHub release badge does not work (first badge in Readme). We can use tag badge instead.


platan:
I agree with your suggestion. It would be great to have releases with notes and binaries (I leave it for the maintainers of this project). I wanted to fix version badge because "release | none" does not look good.

PR #62, opened 2017-11-19
fix: preallocate maps

cristaloleg:


janisz:
It will be cool if we can add benchmark to show this change actually improved performance.

cristaloleg:
ok, will do this

cristaloleg:
Changes are quite neglectable, it's used in cold paths.
Merge it or not is your decision 😉

mjarco:
LGTM. I'll merge after running some more tests. Thanks!

PR #64, opened 2017-11-30
Listing merge

mjarco:


crdstore/crdstore.go
```diff
@@ -95,6 +95,7 @@ func (cs *CredentialsStore) updateCache(accessKey, backend, key string, csd *Cre
 }
 
 func (cs *CredentialsStore) tryLock() bool {
+	// #nosec
 	return atomic.CompareAndSwapInt32((*int32)(unsafe.Pointer(&cs.lock)), 0, 1)
```
janisz: Please add comment why it's done that way

crdstore/crdstore_test.go
```diff
@@ -86,14 +86,14 @@ func TestShouldPrepareInternalKeyBasedOnAccessAndStorageType(t *testing.T) {
...
-	require.Equal(t, existingCredentials.SecretKey, csd.SecretKey, "key must be equal")
-}
+// func TestShouldSetCredentialsFromExternalServiceEndpoint(t *testing.T) {
+// 	cs, _ := GetInstance("default")
+
+// 	csd, err := cs.Get(existingCredentials.AccessKey, existingStorage)
+// 	require.NoError(t, err)
+// 	require.Equal(t, existingCredentials.AccessKey, csd.AccessKey, "key must be equal")
+// 	require.Equal(t, existingCredentials.SecretKey, csd.SecretKey, "key must be equal")
+// }
```
janisz: Delete if not needed or [`t.Skip("Why?")`](https://golang.org/pkg/testing/#T.Skip)

crdstore/crdstore.go
mjarco: We'll get rid of this construct in future version. I promise :)

glide.yaml
```diff
@@ -2,30 +2,27 @@ package: github.com/allegro/akubra
 import:
 - package: github.com/ShowMax/go-fqdn
 - package: github.com/alecthomas/kingpin
-  version: ^2.2.5
```
janisz: Why you are removing versions?

httphandler/httphandler.go
```diff
@@ -67,10 +67,13 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 	defer func() {
-		closeErr := resp.Body.Close()
-		if closeErr != nil {
-			log.Printf("Cannot send response body reason: %q",
-				closeErr.Error())
+		if resp.Body != nil {
+			closeErr := resp.Body.Close()
+			if closeErr != nil {
```
janisz: Prevent nested conditions with reverting it

```go
if resp.Body == nil {
  return
}
if err := resp.Body.Close(); err != nil {
   log.Printf("Cannot send response body reason: %q", err.Error())
}
```

httphandler/httphandler.go
```diff
@@ -80,9 +83,11 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	}
 
 	w.WriteHeader(resp.StatusCode)
-	if _, copyErr := io.Copy(w, resp.Body); copyErr != nil {
-		log.Printf("Cannot send response body reason: %q",
-			copyErr.Error())
+	if resp.Body != nil {
```
janisz: ditto

PR #67, opened 2017-12-06
maintenance-mode-for-backends

bieli:
+ upgrade validator.v1

### DIFF
--- a/sharding/shards_ring.go
+++ b/sharding/shards_ring.go
@@ -109,7 +109,7 @@ func (sr ShardsRing) regressionCall(cl storages.NamedCluster, req *http.Request)
 	// Do regression call if response status is > 400
 	if (err != nil || resp.StatusCode > 400) && req.Method != http.MethodPut {
 		rcl, ok := sr.clusterRegressionMap[cl.Name()]
-		if ok && resp.Body != nil {
+		if ok && resp != nil && resp.Body != nil {
 			_, discardErr := io.Copy(ioutil.Discard, resp.Body)
 			if discardErr != nil {
 				reqID, _ := req.Context().Value(log.ContextreqIDKey).(string)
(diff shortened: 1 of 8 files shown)


storages/storages.go
```diff
@@ -94,11 +98,9 @@ func newCluster(name string, backendNames []string, backends map[string]http.Rou
 	}
 	for _, backendName := range backendNames {
 		backendRT, ok := backends[backendName]
-		if !ok {
-			return nil, fmt.Errorf("no such backend %q in 'storages::newCluster'", backendName)
+		if ok {
+			clusterBackends = append(clusterBackends, backendRT)
 		}
```
mjarco: Add "else block" which returs error as on master. If there is no required backend name registered in backends map it's obviously incorrect setup and we should fail

sharding/shards_ring.go
```diff
@@ -109,7 +109,7 @@ func (sr ShardsRing) regressionCall(cl storages.NamedCluster, req *http.Request)
 	// Do regression call if response status is > 400
 	if (err != nil || resp.StatusCode > 400) && req.Method != http.MethodPut {
 		rcl, ok := sr.clusterRegressionMap[cl.Name()]
-		if ok {
+		if ok && resp != nil && resp.Body != nil {
```
janisz: How about 
```
if !ok {
    continoue;
} 
```

so you do not need the same condition twice

storages/storages.go
```diff
@@ -44,8 +44,9 @@ type Storages struct {
 // Backend represents any storage in akubra cluster
 type Backend struct {
 	http.RoundTripper
-	Endpoint url.URL
-	Name     string
+	Endpoint    url.URL
+	Name        string
+	Maintenance bool
```
janisz: Documentation :exclamation: Add information about this feature in README and godoc here

PR #68, opened 2017-12-08
Slow lint

mjarco:


config/config.go
```diff
@@ -61,7 +61,7 @@ func Configure(configFilePath string) (conf Config, err error) {
 		log.Fatalf("[ ERROR ] Problem with opening config file: '%s' - err: %v !", configFilePath, err)
 		return conf, err
 	}
-	defer confFile.Close()
+	defer func() { err = confFile.Close(); log.Debugf("Cannot close configuration, reason: %s", err) }()
```
janisz: Missing `if`?

```
if err = confFile.Close(); err != nil {
    log.Debugf("Cannot close configuration, reason: %s", err)
}
```

config/config.go
```diff
@@ -113,29 +126,37 @@ func ValidateConfigurationHTTPHandler(w http.ResponseWriter, r *http.Request) {
...
-		io.WriteString(w, fmt.Sprintf("YAML Unmarshal Error: %s", err))
+		_, ioerr := io.WriteString(w, fmt.Sprintf("YAML Unmarshal Error: %s", err))
+		logWriteHeaderErr(ioerr, "bad request")
 		return
 	}
-	defer r.Body.Close()
+
+	defer func() {
+		err := r.Body.Close()
+		if err != nil {
```
janisz:
```
if err := r.Body.Close(); err != nil {
```

storages/response_handler.go
```diff
@@ -122,19 +116,23 @@ func (rm *responseMerger) createResponse(successes []transport.ResErrTuple) (res
 	}
 	oContainer := objectsContainer{
 		list: make([]ObjectInfo, 0),
-		set:  make(map[string]struct{}, 0),
+		set:  make(map[string]struct{}),
```
janisz: Please avoid using plain strings as a key. What are they?. Use type aliasing to give them proper domain name.

https://testing.googleblog.com/2017/11/obsessed-with-primitives.html
http://wiki.c2.com/?PrimitiveObsession
http://wiki.c2.com/?StringlyTyped

PR #69, opened 2017-12-13
Added brimapi log hook

mjarco:


httphandler/response_merger.go
```diff
@@ -36,6 +38,22 @@ func extractDestinationHostName(r transport.ResErrTuple) string {
 	return ""
 }
 
+func extractAccessKey(req *http.Request) string {
```
janisz: Add a comment here. I do understand what is happening but I do not understand why

PR #72, opened 2018-02-01
Fix bucket acl request handling

mjarco:


storages/response_handler.go
```diff
@@ -211,9 +211,10 @@ func (rm *responseMerger) responseHandler(in <-chan transport.ResErrTuple) trans
 	firstTuple := <-in
 	path := firstTuple.Req.URL.Path
 	method := firstTuple.Req.Method
+	reqQuery := firstTuple.Req.URL.Query()
+
+	if (reqQuery != nil && reqQuery["acl"] != nil) || (method != http.MethodGet || !isBucketPath(path)) {
```
ksitak: do we need to have:
 method != http.MethodGet || !isBucketPath(path) 
in separate brackets?

mjarco: nope

PR #76, opened 2018-03-05
V1.0 transports behaviours

bieli:
In configuration we can specifying Triggers for different Transports with timeouts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,6 @@ Service:
      -
        Triggers:
         - Method: GET|POST
-        - Header: Content-Length > 10000
         - Path: /\s+/\s+$
        MergingStrategy: Default
        Details:
@@ -138,7 +137,6 @@ Service:
      -
        Triggers:
         - Method: GET|POST|PUT
-        - Header: Content-Length > 5000000
         - QueryParam: clientId=\s+
        MergingStrategy: ListV1
        Details:
--- a/README.md
+++ b/README.md
@@ -123,9 +123,10 @@ Service:
       - PUT
       - DELETE
     # behaviours in application with HTTP protocol condition for better transmission
-    # DefaultTransport it's only one required item in Transports section (logical validator is implemented)
+    # DefaultTransport (with empty Triggers) it's only one required item in Transports section (logical validator is implemented)
     Transports:
-      Transport1:
+      -
+        Name: Transport1
         Triggers:
           Method: GET|POST
           Path: .*
@@ -134,16 +135,18 @@ Service:
           MaxIdleConnsPerHost: 1000
           IdleConnTimeout: 2s
           ResponseHeaderTimeout: 5s
-      Transport2:
+      -
+        Name: Transport2
         Triggers:
           Method: GET|POST|PUT
-          QueryParam: clientId=\s+
+          QueryParam: acl
         Details:
           MaxIdleConns: 200
           MaxIdleConnsPerHost: 500
           IdleConnTimeout: 5s
           ResponseHeaderTimeout: 5s
-      DefaultTransport:
+      -
+        Name: DefaultTransport
         Triggers:
         Details:
           MaxIdleConns: 500
--- a/README.md
+++ b/README.md
@@ -123,13 +123,12 @@ Service:
       - PUT
       - DELETE
     # behaviours in application with HTTP protocol condition for better transmission
-    # DefaultTransport it's only one required item in Transports section
+    # DefaultTransport it's only one required item in Transports section (logical validator is implemented)
     Transports:
       Transport1:
         Triggers:
           Method: GET|POST
           Path: .*
-        MergingStrategy: Default
         Details:
           MaxIdleConns: 200
           MaxIdleConnsPerHost: 1000
@@ -139,7 +138,6 @@ Service:
         Triggers:
           Method: GET|POST|PUT
           QueryParam: clientId=\s+
-        MergingStrategy: ListV1
         Details:
           MaxIdleConns: 200
           MaxIdleConnsPerHost: 500
@@ -147,7 +145,6 @@ Service:
           ResponseHeaderTimeout: 5s
       DefaultTransport:
         Triggers:
-        MergingStrategy: Default
         Details:
           MaxIdleConns: 500
           MaxIdleConnsPerHost: 500
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ Service:
       - PUT
       - DELETE
     # behaviours in application with HTTP protocol condition for better transmission
+    # DefaultTransport it's only one required item in Transports section
     Transports:
       Transport1:
         Triggers:
--- a/README.md
+++ b/README.md
@@ -124,26 +124,35 @@ Service:
       - DELETE
     # behaviours in application with HTTP protocol condition for better transmission
     Transports:
-     -
-       Triggers:
-        - Method: GET|POST
-        - Path: /\s+/\s+$
-       MergingStrategy: Default
-       Details:
-         - MaxIdleConns: 200
-         - MaxIdleConnsPerHost: 1000
-         - IdleConnTimeout: 2s
-         - ResponseHeaderTimeout: 5s
-     -
-       Triggers:
-        - Method: GET|POST|PUT
-        - QueryParam: clientId=\s+
-       MergingStrategy: ListV1
-       Details:
-         - MaxIdleConns: 200
-         - MaxIdleConnsPerHost: 500
-         - IdleConnTimeout: 5s
-         - ResponseHeaderTimeout: 5s
+      Transport1:
+        Triggers:
+          Method: GET|POST
+          Path: .*
+        MergingStrategy: Default
+        Details:
+          MaxIdleConns: 200
+          MaxIdleConnsPerHost: 1000
+          IdleConnTimeout: 2s
+          ResponseHeaderTimeout: 5s
+      Transport2:
+        Triggers:
+          Method: GET|POST|PUT
+          QueryParam: clientId=\s+
+        MergingStrategy: ListV1
+        Details:
+          MaxIdleConns: 200
+          MaxIdleConnsPerHost: 500
+          IdleConnTimeout: 5s
+          ResponseHeaderTimeout: 5s
+      DefaultTransport:
+        Triggers:
+        MergingStrategy: Default
+        Details:
+          MaxIdleConns: 500
+          MaxIdleConnsPerHost: 500
+          IdleConnTimeout: 2s
+          ResponseHeaderTimeout: 2s
+
 
 # MaxIdleConns see: https://golang.org/pkg/net/http/#Transport
 # Default 0 (no limit)
--- a/README.md
+++ b/README.md
@@ -93,22 +93,60 @@ the pool and error is logged.
 Configuration is read from a YAML configuration file with the following fields:
 
 ```yaml
-# Listen interface and port e.g. "127.0.0.1:9090", ":80"
-Listen: ":8080"
-# Technical endpoint interface
-TechnicalEndpointListen: ":8071"
-# Technical health check endpoint (for load balancers)
-HealthCheckEndpoint: "/status/ping"
-# Additional not AWS S3 specific headers proxy will add to original request
-AdditionalRequestHeaders:
-    'Cache-Control': "public, s-maxage=600, max-age=600"
-    'X-Akubra-Version': '0.9.26'
-# Additional headers added to backend response
-AdditionalResponseHeaders:
-    'Access-Control-Allow-Origin': "*"
-    'Access-Control-Allow-Credentials': "true"
-    'Access-Control-Allow-Methods': "GET, POST, OPTIONS"
-    'Access-Control-Allow-Headers': "DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type"
+Service:
+  Server:
+    BodyMaxSize: 100MB
+    MaxConcurrentRequests: 200
+    # Listen interface and port e.g. "0:8000", "localhost:9090", ":80"
+    Listen: ":7082"
+    # Technical endpoint interface
+    TechnicalEndpointListen: ":7005"
+    # Health check endpoint (for load balancers)
+    HealthCheckEndpoint: "/status/ping"
+  Client:
+    # Additional not AWS S3 specific headers proxy will add to original request
+    AdditionalResponseHeaders:
+        'Access-Control-Allow-Origin': "*"
+        'Access-Control-Allow-Credentials': "true"
+        'Access-Control-Allow-Methods': "GET, POST, OPTIONS"
+        'Access-Control-Allow-Headers': "DNT,X-CustomHeader,Keep-Alive,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,X-CSRFToken"
+        'Cache-Control': "public, s-maxage=600, max-age=600"
+    # Additional headers added to backend response
+    AdditionalRequestHeaders:
+        'Cache-Control': "public, s-maxage=600, max-age=600"
+    # Backends in maintenance mode
+    # MaintainedBackends:
+    #  - "http://s3.dc2.internal"
+    # List request methods to be logged in synclog in case of backend failure
+    SyncLogMethods:
+      - GET
+      - PUT
+      - DELETE
+    # behaviours in application with HTTP protocol condition for better transmission
+    Transports:
+     -
+       Triggers:
+        - Method: GET|POST
+        - Header: Content-Length > 10000
+        - Path: /\s+/\s+$
+       MergingStrategy: Default
+       Details:
+         - MaxIdleConns: 200
+         - MaxIdleConnsPerHost: 1000
+         - IdleConnTimeout: 2s
+         - ResponseHeaderTimeout: 5s
+     -
+       Triggers:
+        - Method: GET|POST|PUT
+        - Header: Content-Length > 5000000
+        - QueryParam: clientId=\s+
+       MergingStrategy: ListV1
+       Details:
+         - MaxIdleConns: 200
+         - MaxIdleConnsPerHost: 500
+         - IdleConnTimeout: 5s
+         - ResponseHeaderTimeout: 5s
+
 # MaxIdleConns see: https://golang.org/pkg/net/http/#Transport
 # Default 0 (no limit)
 MaxIdleConns: 0
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -45,7 +45,6 @@ const (
     Transports:
       DefaultTransport:
         Triggers:
-        MergingStrategy: Default
         Details:
           MaxIdleConns: 500
           MaxIdleConnsPerHost: 500
@@ -411,9 +410,8 @@ func PrepareYamlConfig(
 	if transports == nil {
 		transports = transportConfig.Transports{
 			"TestTransport": transportConfig.Transport{
-				Triggers:        clientTransportTriggers,
-				MergingStrategy: "test",
-				Details:         clientTransportDetail,
+				Triggers: clientTransportTriggers,
+				Details:  clientTransportDetail,
 			},
 		}
 	}
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -13,12 +13,12 @@ import (
 	"strings"
 
 	httpHandlerConfig "github.com/allegro/akubra/httphandler/config"
-	transportConfig "github.com/allegro/akubra/transport/config"
 	logconfig "github.com/allegro/akubra/log/config"
 	"github.com/allegro/akubra/metrics"
 	regionsConfig "github.com/allegro/akubra/regions/config"
 	shardingconfig "github.com/allegro/akubra/sharding/config"
 	storageconfig "github.com/allegro/akubra/storages/config"
+	transportConfig "github.com/allegro/akubra/transport/config"
 	"github.com/stretchr/testify/assert"
 
 	yaml "gopkg.in/yaml.v2"
@@ -42,7 +42,15 @@ const (
       Cache-Control: public, s-maxage=600, max-age=600
     AdditionalResponseHeaders:
       Access-Control-Allow-Methods: GET, POST, OPTIONS
-  
+    Transports:
+      DefaultTransport:
+        Triggers:
+        MergingStrategy: Default
+        Details:
+          MaxIdleConns: 500
+          MaxIdleConnsPerHost: 500
+          IdleConnTimeout: 2s
+          ResponseHeaderTimeout: 2s
 Backends:
   dummy:
     Endpoint: "http://127.0.0.1:8080"
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -402,7 +402,7 @@ func PrepareYamlConfig(
 
 	if transports == nil {
 		transports = transportConfig.Transports{
-			0: transportConfig.Transport{
+			"TestTransport": transportConfig.Transport{
 				Triggers:        clientTransportTriggers,
 				MergingStrategy: "test",
 				Details:         clientTransportDetail,
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -43,7 +43,7 @@ const (
     AdditionalResponseHeaders:
       Access-Control-Allow-Methods: GET, POST, OPTIONS
     Transports:
-      DefaultTransport:
+      - Name: DefaultTransport
         Triggers:
         Details:
           MaxIdleConns: 500
@@ -409,7 +409,8 @@ func PrepareYamlConfig(
 
 	if transports == nil {
 		transports = transportConfig.Transports{
-			"TestTransport": transportConfig.Transport{
+			transportConfig.Transport{
+				Name:     "TestTransport",
 				Triggers: clientTransportTriggers,
 				Details:  clientTransportDetail,
 			},
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	httpHandlerConfig "github.com/allegro/akubra/httphandler/config"
+	transportConfig "github.com/allegro/akubra/transport/config"
 	logconfig "github.com/allegro/akubra/log/config"
 	"github.com/allegro/akubra/metrics"
 	regionsConfig "github.com/allegro/akubra/regions/config"
@@ -70,7 +71,7 @@ func (t *YamlConfigTest) NewYamlConfigTest() *YamlConfig {
 	size.SizeInBytes = 2048
 	region := regionsConfig.Region{}
 	t.YamlConfig = PrepareYamlConfig(size, 31, 45, "127.0.0.1:81", ":80", ":81",
-		map[string]regionsConfig.Region{"region": region})
+		map[string]regionsConfig.Region{"region": region}, nil)
 	return &t.YamlConfig
 }
 
@@ -150,7 +151,7 @@ func TestShouldValidateConfMaintainedBackendWhenNotEmpty(t *testing.T) {
 	size.SizeInBytes = 2048
 	region := regionsConfig.Region{}
 	testConfData := PrepareYamlConfig(size, 21, 32, maintainedBackendHost, ":80", ":81",
-		map[string]regionsConfig.Region{"region": region})
+		map[string]regionsConfig.Region{"region": region}, nil)
 
 	result, _ := ValidateConf(testConfData, false)
 
@@ -164,7 +165,7 @@ func TestShouldValidateConfMaintainedBackendWhenEmpty(t *testing.T) {
 	region := regionsConfig.Region{}
 	testConfData := PrepareYamlConfig(size, 22, 33, maintainedBackendHost,
 		":80", ":81",
-		map[string]regionsConfig.Region{"region": region})
+		map[string]regionsConfig.Region{"region": region}, nil)
 
 	result, _ := ValidateConf(testConfData, false)
 
@@ -337,7 +338,8 @@ func TestShouldNotPassWhenNoRegionIsDefined(t *testing.T) {
 	maintainedBackendHost := "127.0.0.1:85"
 	var size httpHandlerConfig.HumanSizeUnits
 	size.SizeInBytes = 2048
-	testConfData := PrepareYamlConfig(size, 21, 32, maintainedBackendHost, ":80", ":81", regionsConfig.Regions{})
+	testConfData := PrepareYamlConfig(size, 21, 32, maintainedBackendHost,
+		":80", ":81", regionsConfig.Regions{}, nil)
 
 	result, _ := ValidateConf(testConfData, true)
 
@@ -351,7 +353,8 @@ func PrepareYamlConfig(
 	maintainedBackendHost string,
 	listen string,
 	technicalEndpointListen string,
-	regions regionsConfig.Regions) YamlConfig {
+	regions regionsConfig.Regions,
+	transports transportConfig.Transports) YamlConfig {
 
 	url1 := url.URL{Scheme: "http", Host: "127.0.0.1:8080"}
 	yamlURL := shardingconfig.YAMLUrl{URL: &url1}
@@ -383,6 +386,30 @@ func PrepareYamlConfig(
 		"Access-Control-Allow-Methods": "GET, POST, OPTIONS",
 	}
 
+	clientTransportTriggers := transportConfig.ClientTransportTriggers{
+		Method:     "GET",
+		Path:       "/path",
+		QueryParam: "?acl",
+	}
+
+	clientTransportDetail := transportConfig.ClientTransportDetail{
+		MaxIdleConns:          maxIdleConns,
+		MaxIdleConnsPerHost:   maxIdleConnsPerHost,
+		IdleConnTimeout:       metrics.Interval{Duration: idleConnTimeoutInp},
+		ResponseHeaderTimeout: metrics.Interval{Duration: responseHeaderTimeoutInp},
+		DisableKeepAlives:     false,
+	}
+
+	if transports == nil {
+		transports = transportConfig.Transports{
+			0: transportConfig.Transport{
+				Triggers:        clientTransportTriggers,
+				MergingStrategy: "test",
+				Details:         clientTransportDetail,
+			},
+		}
+	}
+
 	return YamlConfig{
 		Service: httpHandlerConfig.Service{
 			Server: httpHandlerConfig.Server{
@@ -393,13 +420,11 @@ func PrepareYamlConfig(
 				BodyMaxSize:             bodyMaxSize,
 			},
 			Client: httpHandlerConfig.Client{
-				MaxIdleConns:              maxIdleConns,
-				MaxIdleConnsPerHost:       maxIdleConnsPerHost,
-				IdleConnTimeout:           metrics.Interval{Duration: idleConnTimeoutInp},
-				ResponseHeaderTimeout:     metrics.Interval{Duration: responseHeaderTimeoutInp},
-				AdditionalRequestHeaders:  additionalRequestHeaders,
-				AdditionalResponseHeaders: additionalResponseHeaders,
-				DisableKeepAlives:         false,
+				Transports: transports,
+				ClientAdditionalHeader: httpHandlerConfig.ClientAdditionalHeader{
+					AdditionalRequestHeaders:  additionalRequestHeaders,
+					AdditionalResponseHeaders: additionalResponseHeaders,
+				},
 			},
 		},
 
--- a/config/validator.go
+++ b/config/validator.go
@@ -89,14 +89,29 @@ func (c *YamlConfig) RegionsEntryLogicalValidator() (valid bool, validationError
 	for regionName, regionConf := range c.Regions {
 		errList = append(errList, c.validateRegionCluster(regionName, regionConf)...)
 	}
-	if len(errList) > 0 {
-		valid = false
-		errorsList := make(map[string][]error)
-		errorsList["RegionsEntryLogicalValidator"] = errList
-		validationErrors = mergeErrors(validationErrors, errorsList)
-	} else {
-		valid = true
+	validationErrors, valid = prepareErrors(errList, "RegionsEntryLogicalValidator")
+	return
+}
+
+// TransportsEntryLogicalValidator checks the correctness of "Transports" part of configuration file
+func (c *YamlConfig) TransportsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
+	errList := make([]error, 0)
+	if len(c.Service.Transports) == 0 {
+		errList = append(errList, errors.New("Empty transports definition"))
 	}
+	lastTransportHasEmptyTriggers := false
+	for _, transportConf := range c.Service.Transports {
+		methodIsDeclared, pathIsDeclared, queryIsDeclared :=
+			len(transportConf.Triggers.Method) > 0, len(transportConf.Triggers.Path) > 0, len(transportConf.Triggers.QueryParam) > 0
+
+		if methodIsDeclared && pathIsDeclared && queryIsDeclared {
+			lastTransportHasEmptyTriggers = true
+		}
+	}
+	if !lastTransportHasEmptyTriggers {
+		errList = append(errList, errors.New("No transport defined with empty \"Triggers\" in last item (dafault transport)"))
+	}
+	validationErrors, valid = prepareErrors(errList, "TransportsEntryLogicalValidator")
 	return
 }
 
@@ -129,6 +144,18 @@ func mergeErrors(maps ...map[string][]error) (output map[string][]error) {
 		}
 	}
 	return output
+}
+
+// prepareErrors
+func prepareErrors(errList []error, validatorName string) (validationErrors map[string][]error, valid bool) {
+	valid = true
+	if len(errList) > 0 {
+		valid = false
+		errorsList := make(map[string][]error)
+		errorsList[validatorName] = errList
+		validationErrors = mergeErrors(validationErrors, errorsList)
+	}
+	return
 }
 
 // RequestHeaderContentTypeValidator for Content-Type header in request
--- a/config/validator.go
+++ b/config/validator.go
@@ -96,20 +96,21 @@ func (c *YamlConfig) RegionsEntryLogicalValidator() (valid bool, validationError
 // TransportsEntryLogicalValidator checks the correctness of "Transports" part of configuration file
 func (c *YamlConfig) TransportsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
 	errList := make([]error, 0)
-	if len(c.Service.Transports) == 0 {
+	if len(c.Service.Client.Transports) == 0 {
 		errList = append(errList, errors.New("Empty transports definition"))
-	}
-	lastTransportHasEmptyTriggers := false
-	for _, transportConf := range c.Service.Transports {
-		methodIsDeclared, pathIsDeclared, queryIsDeclared :=
-			len(transportConf.Triggers.Method) > 0, len(transportConf.Triggers.Path) > 0, len(transportConf.Triggers.QueryParam) > 0
+	} else {
+		lastTransportHasEmptyTriggers := false
+		for _, transportConf := range c.Service.Client.Transports {
+			methodIsDeclared, pathIsDeclared, queryIsDeclared :=
+				len(transportConf.Triggers.Method) > 0, len(transportConf.Triggers.Path) > 0, len(transportConf.Triggers.QueryParam) > 0
 
-		if methodIsDeclared && pathIsDeclared && queryIsDeclared {
-			lastTransportHasEmptyTriggers = true
+			if !methodIsDeclared && !pathIsDeclared && !queryIsDeclared {
+				lastTransportHasEmptyTriggers = true
+			}
 		}
-	}
-	if !lastTransportHasEmptyTriggers {
-		errList = append(errList, errors.New("No transport defined with empty \"Triggers\" in last item (dafault transport)"))
+		if !lastTransportHasEmptyTriggers {
+			errList = append(errList, errors.New("No transport defined with empty \"Triggers\" in last item (dafault transport)"))
+		}
 	}
 	validationErrors, valid = prepareErrors(errList, "TransportsEntryLogicalValidator")
 	return
--- a/config/validator_test.go
+++ b/config/validator_test.go
@@ -14,20 +14,6 @@ import (
 	validator "gopkg.in/validator.v1"
 )
 
-// import (
-// 	"testing"
-
-// 	"net/http"
-// 	"net/http/httptest"
-
-// 	shardingconfig "github.com/allegro/akubra/sharding/config"
-// 	validator "gopkg.in/validator.v1"
-
-// 	"errors"
-
-// 	"github.com/stretchr/testify/assert"
-// )
-
 type CustomItemsTestUnique struct {
 	Items []string `validate:"UniqueValuesSlice=Items"`
 }
@@ -290,11 +276,9 @@ func TestValidatorShouldProcessTransportsWithSuccess(t *testing.T) {
 			Triggers: transportconfig.ClientTransportTriggers{
 				Path: ".*",
 			},
-			MergingStrategy: "test",
 		},
 		"DefaultTransport": transportconfig.Transport{
 			Triggers:        transportconfig.ClientTransportTriggers{},
-			MergingStrategy: "Default",
 		},
 	}
 	var size httphandlerconfig.HumanSizeUnits
@@ -311,7 +295,6 @@ func TestValidatorShouldFailWithMissingLastTransportsItemWithoutTriggerDefinitio
 			Triggers: transportconfig.ClientTransportTriggers{
 				Method: "GET",
 			},
-			MergingStrategy: "test",
 		},
 	}
 	var size httphandlerconfig.HumanSizeUnits
--- a/config/validator_test.go
+++ b/config/validator_test.go
@@ -91,7 +91,7 @@ func TestShouldPassListenPortsLogicalValidator(t *testing.T) {
 	regionConfig := regionsconfig.Region{}
 	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81", listen,
 		listenTechnicalEndpoint,
-		map[string]regionsconfig.Region{"region": regionConfig})
+		map[string]regionsconfig.Region{"region": regionConfig}, nil)
 	valid, validationErrors := yamlConfig.ListenPortsLogicalValidator()
 
 	assert.Len(t, validationErrors, 0, "Should not be errors")
@@ -106,7 +106,7 @@ func TestShouldNotPassListenPortsLogicalValidatorWhenPortsAreEqual(t *testing.T)
 	regionConfig := regionsconfig.Region{}
 	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81", listen,
 		listenTechnicalEndpoint,
-		map[string]regionsconfig.Region{"region": regionConfig})
+		map[string]regionsconfig.Region{"region": regionConfig}, nil)
 	valid, validationErrors := yamlConfig.ListenPortsLogicalValidator()
 
 	assert.Len(t, validationErrors, 1, "Should be one error")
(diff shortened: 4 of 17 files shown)


transport/transport.go
```diff
@@ -306,3 +319,55 @@ func NewMultiTransport(backends []http.RoundTripper,
 		Backends:        backends,
 		HandleResponses: responsesHandler}
 }
+
+// SetTransportsConfig assign transport config to Container
+func (tc *Container) SetTransportsConfig(clientConfig httphandlerConfig.Client) {
+	tc.TransportsConfig = clientConfig.Transports
+}
+
+// SelectTransport return transport name by method, path and queryParams
```
ksitak: returns

transport/transport.go
```diff
@@ -306,3 +319,55 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+func (tc *Container) SelectTransport(method, path, queryParams string) (transportName string) {
+	_, transportName, ok := tc.TransportsConfig.GetMatchedTransport(method, path, queryParams)
+	if !ok {
+		log.DefaultLogger.Fatalf("Transport not matched with args. method: %s, path: %s, queryParams: %s", method, path, queryParams)
+	}
+	return
+}
+
+// ConfigureHTTPTransportsContainer returns map Transports names from config. with http.Transport with customized dialer
```
ksitak: what dot after config means?

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+		if methodFlag.empty && pathFlag.empty && queryParamFlag.empty && len(defaultTransportName) == 0 {
+			defaultTransport = transport
+			defaultTransportName = transport.Name
+			ok = true
+		}
+	}
+	return
+}
+
+// matchTransportFlags matching method, path and query for Transport
```
ksitak: not matches?

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+			if err != nil {
+				return fmt.Errorf("compileRule for Client->Transport->Trigger->QueryParam error: %q", err)
+			}
+		}
+		t.TriggersCompiledRules.IsCompiled = true
+	}
+	return nil
+}
+
+// GetMatchedTransport return first details matching with rules from Triggers by arguments: method, path, queryParam
```
ksitak: returns

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// transportFlags for internal matching func
+type transportFlags struct {
+	declared bool
+	matched  bool
+	empty    bool
+}
+
+// compileRules preparing precompiled regular expressions for rules
```
ksitak: prepares

storages/backend_test.go
```diff
@@ -1,22 +1,33 @@
...
 
+//
+//type transportContainerRt struct {
+//	rt func(*http.Request) (*http.Response, error)
+//}
+//
+//func (trt *transportContainerRt) RoundTrip(req *http.Request) (*http.Response, error) {
+//	return trt.rt(req)
+//}
+/*
```
ksitak: ?

config/validator_test.go
```diff
@@ -258,11 +246,68 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
+		"127.0.0.1:1234", "127.0.0.1:1235", regions, nil)
 	valid, validationErrors := yamlConfig.RegionsEntryLogicalValidator()
 	assert.False(t, valid)
 	assert.Equal(
 		t,
 		errors.New("No clusters defined for region \"testregion\""),
 		validationErrors["RegionsEntryLogicalValidator"][0])
 }
+
```
wookie41: Shouldn't there be a test for a case, when a default Transport definition is missing?

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+			if err != nil {
+				return fmt.Errorf("compileRule for Client->Transport->Trigger->Path error: %q", err)
+			}
+		}
+		if len(t.Triggers.QueryParam) > 0 {
+			var err error
+			t.TriggersCompiledRules.QueryParamRegexp, err = t.compileRule(t.Triggers.QueryParam)
+			if err != nil {
+				return fmt.Errorf("compileRule for Client->Transport->Trigger->QueryParam error: %q", err)
+			}
```
wookie41: Maybe we can replace the repetitive code with some smart loop?

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		t.TriggersCompiledRules.IsCompiled = true
+	}
+	return nil
+}
+
+// GetMatchedTransport returns first details matching with rules from Triggers by arguments: method, path, queryParam
+func (t *Transports) GetMatchedTransport(method, path, queryParam string) (defaultTransport Transport, defaultTransportName string, ok bool) {
+	for _, transport := range *t {
+		transport.compileRules()
```
wookie41: Unhandled error

config/validator.go
```diff
@@ -89,14 +89,30 @@ func (c *YamlConfig) RegionsEntryLogicalValidator() (valid bool, validationError
...
+	if len(c.Service.Client.Transports) == 0 {
+		errList = append(errList, errors.New("Empty transports definition"))
 	} else {
-		valid = true
+		lastTransportHasEmptyTriggers := false
+		for _, transportConf := range c.Service.Client.Transports {
+			methodIsDeclared, pathIsDeclared, queryIsDeclared :=
+				len(transportConf.Triggers.Method) > 0, len(transportConf.Triggers.Path) > 0, len(transportConf.Triggers.QueryParam) > 0
+
+			if !methodIsDeclared && !pathIsDeclared && !queryIsDeclared {
```
wookie41:
```
if !methodIsDeclared && !pathIsDeclared && !queryIsDeclared {
   errList = append(errList, errors.New("No transport defined with empty \"Triggers\" in last item (dafault transport)"))
   break
} 
```
?

config/validator.go
```diff
@@ -131,6 +147,18 @@ func mergeErrors(maps ...map[string][]error) (output map[string][]error) {
...
+func prepareErrors(errList []error, validatorName string) (validationErrors map[string][]error, valid bool) {
+	valid = true
+	if len(errList) > 0 {
+		valid = false
+		errorsList := make(map[string][]error)
+		errorsList[validatorName] = errList
+		validationErrors = mergeErrors(validationErrors, errorsList)
+	}
+	return
+}
```
wookie41:
```
if valid = len(errlist) < 1; !valid { 
...
}
```

storages/multipart_round_tripper_test.go
```diff
@@ -335,3 +351,20 @@ func testBadResponse(statusCode int, xmlResponse string, testSuite *testing.T) {
 	activeBackendRoundTripper1.AssertNumberOfCalls(testSuite, "RoundTrip", 1)
 	syncLog.AssertNumberOfCalls(testSuite, "Println", 0)
 }
+
+func preapreTestContainerByRoundTripper(roundTripper http.RoundTripper) transport.Container {
```
wookie41: typo in `prepare`

storages/storages.go
```diff
@@ -61,10 +61,11 @@ type Storages struct {
 
 // Backend represents any storage in akubra cluster
 type Backend struct {
-	http.RoundTripper
-	Endpoint    url.URL
-	Name        string
-	Maintenance bool
+	Transports   transport.Container
+	RoundTripper http.RoundTripper
```
wookie41: Why do we still have a seperate RoundTripper, when we already have `Transports` with multiple RoundTrippers?

transport/transport.go
```diff
@@ -306,3 +320,58 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+	maxIdleConnsPerHost := defaultMaxIdleConnsPerHost
+	responseHeaderTimeout := defaultResponseHeaderTimeout
+	if len(clientConf.Transports) > 0 {
+		var defaultRoundTripper http.RoundTripper
+		for _, transport := range clientConf.Transports {
+			roundTrippers[transport.Name] = perepareTransport(transport.Details, maxIdleConnsPerHost, responseHeaderTimeout)
+			defaultRoundTripper = roundTrippers[transport.Name]
+		}
+		transportContainer.RoundTrippers = roundTrippers
+		transportContainer.DefaultRoundTripper = defaultRoundTripper
```
wookie41: How do we know that the chosen transport can be treated as a default on in every case? Shouldn't the default transport be the `DefaultTransport` from config file?

transport/transport.go
```diff
@@ -55,6 +62,13 @@ func (r *ResErrTuple) DiscardBody() {
 // returned value's response and error will be passed to client
 type MultipleResponsesHandler func(in <-chan ResErrTuple) ResErrTuple
 
+// Container mapping initialized Transports with http.RoundTripper by transport name
+type Container struct {
+	DefaultRoundTripper http.RoundTripper
```
wookie41: We can define a constant `defaultTransportName` and just use it on `RoundTrippers` map

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+		if methodFlag.matched && pathFlag.matched && queryParamFlag.matched {
+			return transport, transport.Name, true
+		}
+		if methodFlag.empty && pathFlag.empty && queryParamFlag.empty && len(defaultTransportName) == 0 {
+			defaultTransport = transport
+			defaultTransportName = transport.Name
+			ok = true
+		}
+	}
+	return
```
wookie41: What for the `ok` return value, if it's always `true`?

transport/config/config.go
```diff
@@ -0,0 +1,143 @@
...
+				return fmt.Errorf("compileRule for Client->Transport->Trigger->QueryParam error: %q", err)
+			}
+		}
+		t.TriggersCompiledRules.IsCompiled = true
+	}
+	return nil
+}
+
+// GetMatchedTransport returns first details matching with rules from Triggers by arguments: method, path, queryParam
+func (t *Transports) GetMatchedTransport(method, path, queryParam string) (defaultTransport Transport, defaultTransportName string, ok bool) {
```
wookie41: shouldn't `defaultTransport` and `defaulTransportName` be `matchedTransport` and `matchedTransportName`?

transport/transport.go
```diff
@@ -306,3 +320,58 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+
+// SetTransportsConfig assign transport config to Container
+func (tc *Container) SetTransportsConfig(clientConfig httphandlerConfig.Client) {
+	tc.TransportsConfig = clientConfig.Transports
+}
+
+// SelectTransport returns transport name by method, path and queryParams
+func (tc *Container) SelectTransport(method, path, queryParams string) (transportName string) {
+	_, transportName, ok := tc.TransportsConfig.GetMatchedTransport(method, path, queryParams)
+	if !ok {
```
wookie41: IMHO this branch is not necessary, as we can alwasy pick a transport for request (if none matches - pick the default)

README.md
```diff
@@ -93,22 +93,68 @@ the pool and error is logged.
...
+      - PUT
+      - DELETE
+    # behaviours in application with HTTP protocol condition for better transmission
+    # DefaultTransport (with empty Triggers) it's only one required item in Transports section (logical validator is implemented)
+    Transports:
+      -
+        Name: Transport1
+        Triggers:
+          Method: GET|POST
+          Path: .*
```
wookie41: when I provide an empty `Path`, `Name`, or `Method` akubra starts, but crashes on serving request.
when I provide a following config 
```Transports:
      -
        Name:
        Triggers:
          Method:
          Path:
        Details:
          MaxIdleConns: 200
          MaxIdleConnsPerHost: 1000
          IdleConnTimeout: 2s
          ResponseHeaderTimeout: 5s
```
It starts and servers requests, even tho no proper transport is defined. Why?

transport/config/config.go
bieli: Maybe, but this implementation is readable in my opinion.
Alternatively to implementing loop for this we need reflection or return values by reference. It will be more complicated for code reading, I think.

README.md
```diff
@@ -93,22 +93,68 @@ the pool and error is logged.
...
+      - GET
+      - PUT
+      - DELETE
+    # behaviours in application with HTTP protocol condition for better transmission
+    # DefaultTransport (with empty Triggers) it's only one required item in Transports section (logical validator is implemented)
+    Transports:
+      -
+        Name: Transport1
+        Triggers:
+          Method: GET|POST
```
wookie41: When I provide an invalid separator akubra starts but crashes on request

transport/config/config.go
bieli: Done

config/validator.go
bieli: Below suggestion doesn't work without additional condition.
Done (removing few lines of code).

config/validator.go
bieli: Done

storages/multipart_round_tripper_test.go
bieli: Done

transport/config/config.go
bieli: Done

transport/config/config.go
bieli: Done

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+
+// Transports map with Transport
+type Transports []Transport
+
+// compileRule
+func (t *Transport) compileRule(regexpRule string) (compiledRule *regexp.Regexp, err error) {
+	if len(regexpRule) > 0 {
+		compiledRule, err = regexp.Compile(regexpRule)
+	}
+	return
```
wookie41: imho an error should be returned if `len(regexpRule)` < 1 - it should make sense to allow for an empty rule

transport/transport.go
bieli: Done

storages/storages.go
```diff
@@ -79,7 +80,12 @@ func (b *Backend) RoundTrip(r *http.Request) (*http.Response, error) {
 			origErr: fmt.Errorf("backend %v in maintenance mode", b.Name)}
 	}
 	err := error(nil)
-	resp, oerror := b.RoundTripper.RoundTrip(r)
+
+	transportName := b.Transports.SelectTransport(r.Method, r.URL.Path, r.URL.RawQuery)
```
mjarco: this should be encapsulated in Transports.RoundTripp method and as Lukasz suggested put transports into roundtripper field

transport/config/config.go
bieli: As you can see it doesn't compile empty rules. All empty rules in section `Matchers` are allow and akubra selecting `default Transport`.
If we allow empty rules in `Matchers` it's impossible to throws Error in one rule (e.g. in `else`) with current specification.

transport/transport.go
```diff
@@ -55,6 +62,13 @@ func (r *ResErrTuple) DiscardBody() {
 // returned value's response and error will be passed to client
 type MultipleResponsesHandler func(in <-chan ResErrTuple) ResErrTuple
 
+// Container mapping initialized Transports with http.RoundTripper by transport name
+type Container struct {
```
mjarco: 1) Add RoundTrip method with body from backend.RoundTrip as I suggested in storages package
2) Rename it Container is too generic name. Suggestions: Selector, TransportMatcher. Feel free to make your own name which suggests purpose of this struct

storages/storages.go
```diff
@@ -52,19 +52,40 @@ type Cluster struct {
...
-	Maintenance bool
+	RoundTripper http.RoundTripper
+	Endpoint     url.URL
+	Name         string
+	Maintenance  bool
+}
+
+// RoundTrip extends TransportRoundTripper struct wtih RoundTripper and transports container
+func (trt *TransportRoundTripper) RoundTrip(request *http.Request) (*http.Response, error) {
+	return trt.SelectTransportByRequest(request).RoundTrip(request)
```
mjarco: why not trt.RoundTripper.RoundTrip?

storages/storages.go
```diff
@@ -52,19 +52,40 @@ type Cluster struct {
...
-	transport        http.RoundTripper
+	Transports       transport.Container
 	Clusters         map[string]NamedCluster
 	Backends         map[string]http.RoundTripper
 	lateRespHandler  transport.MultipleResponsesHandler
 	earlyRespHandler transport.MultipleResponsesHandler
 }
 
+// TransportRoundTripper for slecte
+type TransportRoundTripper struct {
```
mjarco: Why here? move it to transport package. Acctually don't move it, just implement RoundTrip method within transport Container. This package don't need to know anything about Container internals

storages/storages.go
```diff
@@ -165,7 +186,7 @@ func (st *Storages) ClusterShards(name string, syncLog log.Logger, clusters ...N
 }
 
 // InitStorages setups storages
-func InitStorages(transport http.RoundTripper, clustersConf config.ClustersMap, backendsConf config.BackendsMap, earlyRespHandler, lateRespHandler transport.MultipleResponsesHandler, syncLog log.Logger) (*Storages, error) {
+func InitStorages(transportContainer transport.Container, clustersConf config.ClustersMap, backendsConf config.BackendsMap, earlyRespHandler, lateRespHandler transport.MultipleResponsesHandler, syncLog log.Logger) (*Storages, error) {
```
mjarco: transport http.RoundTripper argument should be left intact if you'll make proposed changes

storages/storages.go
```diff
@@ -194,17 +215,17 @@ func InitStorages(transport http.RoundTripper, clustersConf config.ClustersMap,
...
+		Transports:       transportContainer,
 		Clusters:         clusters,
 		Backends:         backends,
 		earlyRespHandler: earlyRespHandler,
 		lateRespHandler:  lateRespHandler,
 	}, nil
 }
 
-func decorateBackend(transport http.RoundTripper, name string, backendConf config.Backend) (http.RoundTripper, error) {
+func decorateBackend(transports transport.Container, name string, backendConf config.Backend) (http.RoundTripper, error) {
```
mjarco: Again no need to pass specific type

storages/storages.go
```diff
@@ -61,10 +60,10 @@ type Storages struct {
 
 // Backend represents any storage in akubra cluster
 type Backend struct {
-	http.RoundTripper
-	Endpoint    url.URL
-	Name        string
-	Maintenance bool
+	RoundTripper transport.Matcher
```
mjarco: http.RoundTripper

storages/storages.go
```diff
@@ -165,7 +164,7 @@ func (st *Storages) ClusterShards(name string, syncLog log.Logger, clusters ...N
 }
 
 // InitStorages setups storages
-func InitStorages(transport http.RoundTripper, clustersConf config.ClustersMap, backendsConf config.BackendsMap, earlyRespHandler, lateRespHandler transport.MultipleResponsesHandler, syncLog log.Logger) (*Storages, error) {
+func InitStorages(transportMatcher transport.Matcher, clustersConf config.ClustersMap, backendsConf config.BackendsMap, earlyRespHandler, lateRespHandler transport.MultipleResponsesHandler, syncLog log.Logger) (*Storages, error) {
```
mjarco: func InitStorages(http.RoundTripper [...]

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// matchTransportFlags matches method, path and query for Transport
+func matchTransportFlags(transport Transport, method, path, queryParam string) (transportFlags, transportFlags, transportFlags) {
+	var methodFlag, pathFlag, queryParamFlag transportFlags
+
+	methodFlag.declared, pathFlag.declared, queryParamFlag.declared =
+		len(transport.Matchers.Method) > 0, len(transport.Matchers.Path) > 0, len(transport.Matchers.QueryParam) > 0
+
```
mjarco: Readability counts! 
```
methodFlag.declared = len(transport.Matchers.Method) > 0
pathFlag.declared = len(transport.Matchers.Path) > 0
queryParamFlag.declared = len(transport.Matchers.QueryParam) > 0
```

storages/storages.go
bieli: transport.Matcher type in this case is required for successful compilation.
We no need type casting in this implementation.

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+// transportFlags for internal matching func
+type transportFlags struct {
+	declared bool
+	matched  bool
+	empty    bool
+}
+
+// compileRules prepares precompiled regular expressions for rules
+func (t *Transport) compileRules() error {
+	if !t.MatchersCompiledRules.IsCompiled {
```
mjarco: Readability:
```
if t.MatchersCompiledRules.IsCompiled {
    return nil
}
```
And reduce one indentation

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+		if len(t.Matchers.Method) > 0 {
```
mjarco: Readability: You're checking same thing inside `compileRule` method, there is no need to have this if's here

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+			var err error
```
mjarco: no need for explicit deklaration use `:=` operator line below

mjarco: sorry that would not compile

transport/config/config.go
mjarco: BTW, is this sentinel really necessary?

transport/transport.go
```diff
@@ -306,3 +319,71 @@ func NewMultiTransport(backends []http.RoundTripper,
...
 		HandleResponses: responsesHandler}
 }
+
+// SetTransportsConfig assign transport config to Matcher
+func (m *Matcher) SetTransportsConfig(clientConfig httphandlerConfig.Client) {
+	m.TransportsConfig = clientConfig.Transports
+}
+
+// SelectTransport returns transport name by method, path and queryParams
+func (m *Matcher) SelectTransport(method, path, queryParams string) (transportName string) {
```
mjarco: Readability: Missleading method name it does not return transport, but transport name thus method should be called SelectTransportName

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+				return fmt.Errorf("compileRule for Client->Transport->Trigger->QueryParam error: %q", err)
+			}
+		}
+		t.MatchersCompiledRules.IsCompiled = true
+	}
+	return nil
+}
+
+// GetMatchedTransport returns first details matching with rules from Matchers by arguments: method, path, queryParam
+func (t *Transports) GetMatchedTransport(method, path, queryParam string) (matchedTransport Transport, matchedTransportName string, ok bool) {
```
mjarco: None of callers use first returned value, maybe it should not be returrned at all?

transport/transport.go
```diff
@@ -306,3 +319,71 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+}
+
+// SelectTransportByRequest for selecting RoundTripper by request object from transports container
+func (m *Matcher) SelectTransportByRequest(request *http.Request) (selectedRoundTripper http.RoundTripper) {
+	selectedTransportName := m.SelectTransport(request.Method, request.URL.Path, request.URL.RawQuery)
+	reqID := request.Context().Value(log.ContextreqIDKey)
+	log.DefaultLogger.Debugf("Request %s - selected transport name: %s (by method: %s, path: %s, queryParams: %s)",
+		reqID, selectedTransportName, request.Method, request.URL.Path, request.URL.RawQuery)
+
+	return m.RoundTrippers[selectedTransportName]
```
mjarco: I don't understand why SelectTransport does return transport name instead of transport instance. If would do the latter this mapping call could be omitted. Transport name is one of it's properties

transport/transport.go
```diff
@@ -306,3 +319,71 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+}
+
+// ConfigureHTTPTransports returns RoundTrippers mapped by transport name from configuration
+func ConfigureHTTPTransports(clientConf httphandlerConfig.Client) (http.RoundTripper, error) {
+	roundTrippers := make(map[string]http.RoundTripper)
+	transportMatcher := &Matcher{}
+	transportMatcher.SetTransportsConfig(clientConf)
```
mjarco: Remove SetTransportsConfig method, and use 
```
transportMatcher := &Matcher{TransportsConfig: clientConf}
```
here instead

transport/transport.go
```diff
@@ -306,3 +319,71 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+}
+
+// SelectTransportRoundTripper for selecting RoundTripper by request object from transports matcher
+func (m *Matcher) SelectTransportRoundTripper(request *http.Request) (selectedRoundTripper http.RoundTripper) {
+	selectedTransport := m.SelectTransport(request.Method, request.URL.Path, request.URL.RawQuery)
+	reqID := request.Context().Value(log.ContextreqIDKey)
+	log.DefaultLogger.Debugf("Request %s - selected transport name: %s (by method: %s, path: %s, queryParams: %s)",
+		reqID, selectedTransport.Name, request.Method, request.URL.Path, request.URL.RawQuery)
+
+	return m.RoundTrippers[selectedTransport.Name]
```
mjarco: Does this element differ from `selectedTransport`?

mjarco: If not, we should consider removing Matcher.RoundTrippers map

mjarco: Ok. I see where the problem is... it's the name of variable. For readability sakes: rename config.Transport to something that would not be confused with http.Transport and rename methods accordingly

bieli: @mjarco Proposals: `TransportProperties`, `TransportClientProps`, `HttpTransportProps` ... which name is the best for you ?

mjarco: `Kind`

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	QueryParamRegexp *regexp.Regexp
+	IsCompiled       bool
+}
+
+// Transport properties
+type Transport struct {
+	Name                  string                  `yaml:"Name"`
+	Matchers              ClientTransportMatchers `yaml:"Matchers"`
+	MatchersCompiledRules MatchersCompiledRules
+	Details               ClientTransportDetail `yaml:"Details"`
```
mjarco: `Properties`

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	MethodRegexp     *regexp.Regexp
+	PathRegexp       *regexp.Regexp
+	QueryParamRegexp *regexp.Regexp
+	IsCompiled       bool
+}
+
+// Transport properties
+type Transport struct {
+	Name                  string                  `yaml:"Name"`
+	Matchers              ClientTransportMatchers `yaml:"Matchers"`
```
mjarco: `ApplyRules` or smth in common

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	MatchersCompiledRules MatchersCompiledRules
```
mjarco: `CompiledApplyRules`

config/validator.go
```diff
@@ -89,14 +89,17 @@ func (c *YamlConfig) RegionsEntryLogicalValidator() (valid bool, validationError
...
-	} else {
-		valid = true
+	validationErrors, valid = prepareErrors(errList, "RegionsEntryLogicalValidator")
+	return
+}
+
+// TransportsEntryLogicalValidator checks the correctness of "Transports" part of configuration file
+func (c *YamlConfig) TransportsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
+	errList := make([]error, 0)
+	if len(c.Service.Client.Transports) == 0 {
```
wookie41: In case some magic happens, I would suggest to write the condition as follows `if len(c.Service.Client.Transports) < 1`

config/validator.go
```diff
@@ -131,6 +134,16 @@ func mergeErrors(maps ...map[string][]error) (output map[string][]error) {
 	return output
 }
 
+// prepareErrors
+func prepareErrors(errList []error, validatorName string) (validationErrors map[string][]error, valid bool) {
+	if valid = len(errList) < 1; !valid {
+		errorsList := make(map[string][]error)
+		errorsList[validatorName] = errList
+		validationErrors = mergeErrors(validationErrors, errorsList)
```
wookie41: Shouldn't this instruction be outside of the condition? Because if there are no errors, `validationErrors` will be nil? Right?

config/validator_test.go
```diff
@@ -258,11 +246,106 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	}
+	var size httphandlerconfig.HumanSizeUnits
+	size.SizeInBytes = 2048
+	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
+		"127.0.0.1:1234", "127.0.0.1:1235", nil, invalidTransports)
+	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
+	assert.True(t, valid)
+}
+
+func TestShouldValidWithEmptyPropertiesInmatchersDefinition(t *testing.T) {
```
wookie41: Typo `Inmatchers` -> `InMatchers`

config/validator_test.go
```diff
@@ -258,11 +246,106 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	invalidTransports := transportconfig.Transports{
+		transportconfig.Transport{
+			Name: "TestTransport123",
```
wookie41: Maybe the transport's name should be something like `CatchAllRequestsTransport` so people are aware of such a funcrionality

main.go
```diff
@@ -56,6 +57,13 @@ func main() {
 		log.Fatalf("Improperly configured %s", err)
 	}
 
+	valid, errs := config.ValidateConf(conf.YamlConfig, true)
+	if !valid {
+		fmt.Printf("YAML validation - errors: %q", errs)
+		os.Exit(2)
```
wookie41: Maybe instead of magic values use a constant like `const INVALID_CONFIG_ERROR = 2`, and then you can write a more self-describing code, like  `os.Exit(INVALID_CONFIG_ERROR)`

storages/storages.go
```diff
@@ -61,10 +60,10 @@ type Storages struct {
 
 // Backend represents any storage in akubra cluster
 type Backend struct {
-	http.RoundTripper
-	Endpoint    url.URL
-	Name        string
-	Maintenance bool
+	RoundTripper http.RoundTripper
```
wookie41: `http.RoundTripper` is enough

transport/transport.go
wookie41: `TransportRules`?

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+				return fmt.Errorf("compileRule for Client->Transport->Trigger->QueryParam error: %q", err)
+			}
+		}
+		t.MatchersCompiledRules.IsCompiled = true
+	}
+	return nil
+}
+
+// GetMatchedTransport returns first details matching with rules from Matchers by arguments: method, path, queryParam
+func (t *Transports) GetMatchedTransport(method, path, queryParam string) (matchedTransport Transport, ok bool) {
```
wookie41: Maybe a method name `FindAppropriateTransport` would be better?

transport/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		matchedTransportName = transport.Name
+	}
+	return
+}
+
+// matchTransportFlags matches method, path and query for Transport
+func matchTransportFlags(transport Transport, method, path, queryParam string) (transportFlags, transportFlags, transportFlags) {
+	var methodFlag, pathFlag, queryParamFlag transportFlags
+
```
wookie41: Unnecessary empty line

transport/transport.go
mjarco: apply rules are properties of this type

transport/config/config.go
bieli: Yes, only in first iteration (in loop by transports config) we compiling reg. expr. (it's CPU consuming).

mjarco: Ok let it be, but invert this condition so there will be no unnecessary nesting. Also remeber to remove `if len(t.Matchers.Method) > 0 {` checks in this method because it's checked inside `compileRule` method

bieli: `compileRules` func has own checks for not empty reg. exp. for compiling only filled rules
`matchTransportFlags` func it's simple algo. and we need here the same information
I propose to extract below conditions to private func. `checkEmptyMatchers` if it's still necessary.

mjarco: Ok, so method `compileRule` is useless, because all it does is same pattern length check and single line call to regexCompile...

config/validator.go
bieli: `case some magic happens` - what exactlly ?

wookie41: Idk, it's just a sanity check. Range conditions are imho just "safer" in such cases

transport/transport.go
```diff
@@ -306,3 +319,74 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+
+	return
+}
+
+// ConfigureHTTPTransports returns RoundTrippers mapped by transport name from configuration
+func ConfigureHTTPTransports(clientConf httphandlerConfig.Client) (http.RoundTripper, error) {
+	roundTrippers := make(map[string]http.RoundTripper)
+	transportMatcher := &Matcher{TransportsConfig: clientConf.Transports}
+	maxIdleConnsPerHost := defaultMaxIdleConnsPerHost
+	responseHeaderTimeout := defaultResponseHeaderTimeout
```
wookie41: Should we still use defaults?

config/config_test.go
```diff
@@ -41,7 +42,15 @@ const (
       Cache-Control: public, s-maxage=600, max-age=600
     AdditionalResponseHeaders:
       Access-Control-Allow-Methods: GET, POST, OPTIONS
-  
+    Transports:
+      - Name: DefaultTransport
```
wookie41: I think that a name `DefaultTransport` is misleading, as it suggests that the transport supports all kinds of request, when in fact, it doesn't support a `Head` request.

config/validator.go
```diff
@@ -89,14 +89,25 @@ func (c *YamlConfig) RegionsEntryLogicalValidator() (valid bool, validationError
...
+// TransportsEntryLogicalValidator checks the correctness of "Transports" part of configuration file
+func (c *YamlConfig) TransportsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
+	errList := make([]error, 0)
+	if len(c.Service.Client.Transports) == 0 {
+		errList = append(errList, errors.New("Empty transports definition"))
 	} else {
-		valid = true
+		for _, transportConf := range c.Service.Client.Transports {
+			properties := transportConf.Properties
+			if properties.MaxIdleConns == 0 || properties.MaxIdleConnsPerHost == 0 || properties.ResponseHeaderTimeout.Duration == 0 {
```
wookie41: `< 1` as timeouts can't be negative as well

config/validator.go
bieli: but in this case `len` function never got minus value and for below case `== 0` in condition is enough

transport/config/config.go
```diff
@@ -0,0 +1,147 @@
...
+	Rules         ClientTransportRules `yaml:"Rules"`
+	CompiledRules CompiledRules
+	Properties    ClientTransportProperties `yaml:"Properties"`
+}
+
+// Transports map with Transport
+type Transports []Transport
+
+// compileRule
+func (t *Transport) compileRule(regexpRule string) (compiledRule *regexp.Regexp, err error) {
```
mjarco: Do we need this method? `len(regexpRule) > 0` is checked right before call to this method and all it's left is regexp compilation, with nothing more in it.

mjarco: Even error checking is made outside

transport/config/config.go
```diff
@@ -0,0 +1,147 @@
...
+			return fmt.Errorf("compileRule for Client->TransportMatcherDefinition->Trigger->QueryParam error: %q", err)
+		}
+	}
+	t.CompiledRules.IsCompiled = true
+
+	return nil
+}
+
+// GetMatchedTransport returns first details matching with rules from Rules by arguments: method, path, queryParam
+func (t *Transports) GetMatchedTransport(method, path, queryParam string) (matchedTransport TransportMatcherDefinition, ok bool) {
```
mjarco: GetMatchedTransportDefinition

transport/transport.go
```diff
@@ -306,3 +318,69 @@ func NewMultiTransport(backends []http.RoundTripper,
 		Backends:        backends,
 		HandleResponses: responsesHandler}
 }
+
+// SelectTransport returns transport instance by method, path and queryParams
+func (m *Matcher) SelectTransport(method, path, queryParams string, log log.Logger) (matchedTransport config.TransportMatcherDefinition, err error) {
```
mjarco: SelectTransportMatchDefinition?

main.go
ksitak: Marcin loves returning magic exit codes :) But I'm always against it unless it's documented somewhere. I guess it won't be so just use 1 everywhere.

config/validator_test.go
```diff
@@ -258,11 +259,162 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	var size httphandlerconfig.HumanSizeUnits
+	size.SizeInBytes = 2048
+	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
+		"127.0.0.1:1234", "127.0.0.1:1235", nil, validTransports)
+	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
+	assert.True(t, valid)
+}
+
+func TestValidatorShouldProcessTransportsWithSuccessWithNotDefinedRulesProperties(t *testing.T) {
+	invalidTransports := transportconfig.Transports{
```
wookie41: Why it's called `invalidTransports` when in fact in contains valid transport rules definition?

config/validator_test.go
```diff
@@ -258,11 +259,162 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	var size httphandlerconfig.HumanSizeUnits
+	size.SizeInBytes = 2048
+	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
+		"127.0.0.1:1234", "127.0.0.1:1235", nil, invalidTransports)
+	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
+	assert.True(t, valid)
+}
+
+func TestShouldValidWithEmptyPropertiesInRulesDefinition(t *testing.T) {
+	invalidTransports := transportconfig.Transports{
```
wookie41: Same as above

config/validator_test.go
```diff
@@ -258,11 +259,162 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	var size httphandlerconfig.HumanSizeUnits
+	size.SizeInBytes = 2048
+	yamlConfig := PrepareYamlConfig(size, 51, 55, "127.0.0.1:82",
+		"127.0.0.1:1235", "127.0.0.1:1236", nil, invalidTransports)
+	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
+	assert.True(t, valid)
+}
+
+func TestValidatorShouldValidateTransportsWithEmptyRules(t *testing.T) {
+	validTransports := transportconfig.Transports{
```
wookie41: As above

config/validator_test.go
```diff
@@ -258,11 +259,162 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+		IdleConnTimeout: metrics.Interval{
+			Duration: 0,
+		},
+		ResponseHeaderTimeout: metrics.Interval{
+			Duration: 1,
+		},
+		DisableKeepAlives: false,
+	}
+
+	invalidTransports := transportconfig.Transports{
```
wookie41: As above

README.md
```diff
@@ -240,6 +284,21 @@ Response:
     < Content-Length: 2
     OK
 
+## Transports and Rules with dedicated timeouts
+
+It's transparent feature for end users for guarantee high availability and better transmission.
```
ksitak: Why do we speak about transparency here? Everything in a "client" section is transparent for end user, isn't it?

README.md
```diff
@@ -240,6 +284,21 @@ Response:
...
+
+For example, when one specific HTTP method has lag we can modifying timeouts with special 'Rule'.
```
ksitak: we can set timeouts

README.md
```diff
@@ -240,6 +284,21 @@ Response:
...
+Another example, when user adding biggest chunks by multi upload,
```
ksitak: adds big

README.md
```diff
@@ -240,6 +284,21 @@ Response:
...
+default timeout need to be changing into dedicated 'Transport' with 'Rule' for this case.
```
ksitak: needs to be changed with

transport/transport.go
bieli: SelectTransportDefinition ?

README.md
```diff
@@ -240,6 +284,21 @@ Response:
...
+
+We have 'Rules' for 'Transports' definitions:
+ - required minimum one item in 'Transports' section
+ - required empty or one property (Method, Path, QueryParam) in 'Rules' section
+ - if empty 'Rules' section is defined it will be matching with any requests
```
ksitak: it will match

README.md
```diff
@@ -240,6 +284,21 @@ Response:
...
+ - in runtime no matching request (by 'Rules' props. definition) generating responses with code HTTP 5xx
```
ksitak: when transport cannot be matched, http 500 error code will be send to client.

transport/config/config.go
bieli: GetMatchedTransportDefinition ?

README.md
bieli: Yes, it's redundant part.

transport/transport.go
mjarco: LGTM

main.go
bieli: Yeah, exit codes it's old style ;), but when we running software with `supervisor` it's still informative style (In logs we have "crash" code number).

config/validator.go
bieli: It's no problem - all higher order functions using `valid` returned arg. in this case.
In below example value returned by `validationErrors` isn't usable, when `valid == True`.

main.go
ksitak: It's not old style. And I don't know what's informative in "exit code 2".

wookie41: It shouldn't matter what we are running the software with, you can always do `const BadConfigError = 2` and then `Exit(BadConfigError)`

ksitak: But if we do it that way, then:
- where's documentation for possible error codes?
- are we sure that all possible config errors end up with 2?
- are we sure that not other error case returns 2?
- do we have tests for error codes?

config/validator.go
wookie41: k

main.go
wookie41: It would be helpful during code reviews and reading code in general. If it's defined as a constant then we don't need to guess based on the context what caused the error.

ksitak: I understand, but I feel that this approach does not take error codes seriously. No documentation, no proof they don't repeat, etc. My point of view is that we take it seriously (my questions above) or return 1 everywhere.

wookie41: True dat

README.md
```diff
@@ -286,17 +286,17 @@ Response:
 
 ## Transports and Rules with dedicated timeouts
 
-It's transparent feature for end users for guarantee high availability and better transmission.
+It's guarantee high availability and better transmission.
```
ksitak: What this sentence is supposed to mean?

README.md
```diff
@@ -286,17 +286,17 @@ Response:
...
 
-For example, when one specific HTTP method has lag we can modifying timeouts with special 'Rule'.
-Another example, when user adding biggest chunks by multi upload,
-default timeout need to be changing into dedicated 'Transport' with 'Rule' for this case.
+For example, when one specific HTTP method has lag we can set timeouts with special 'Rule'.
+Another example, when user dds big chunks by multi upload,
```
ksitak: adds

README.md
```diff
@@ -286,17 +286,17 @@ Response:
...
+default timeout needs to be changed with into dedicated 'Transport' with 'Rule' for this case.
```
ksitak: without "into"

config/validator_test.go
```diff
@@ -316,13 +316,13 @@ func TestValidatorShouldProcessTransportsWithSuccessWithNotDefinedRulesPropertie
 	var size httphandlerconfig.HumanSizeUnits
 	size.SizeInBytes = 2048
 	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
-		"127.0.0.1:1234", "127.0.0.1:1235", nil, invalidTransports)
+		"127.0.0.1:1234", "127.0.0.1:1235", nil, transports)
 	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
 	assert.True(t, valid)
 }
 
 func TestShouldValidWithEmptyPropertiesInRulesDefinition(t *testing.T) {
```
ksitak: should validate

config/validator_test.go
```diff
@@ -363,13 +363,13 @@ func TestValidatorShouldValidateTransportsWithEmptyRules(t *testing.T) {
 	var size httphandlerconfig.HumanSizeUnits
 	size.SizeInBytes = 2048
 	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
-		"127.0.0.1:1234", "127.0.0.1:1235", nil, validTransports)
+		"127.0.0.1:1234", "127.0.0.1:1235", nil, transports)
 	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
 	assert.True(t, valid)
 }
 
 func TestShouldFailTransportsEntryLogicalValidatorWithoutProperties(t *testing.T) {
```
ksitak: validator should fail on transport without properties

transport/transport_matcher_test.go
```diff
@@ -0,0 +1,74 @@
...
+		TransportsConfig: clientConfig.Transports,
+	}
+
+	selectedTransport, err := unit.SelectTransportDefinition(testRequest.Method, testRequest.URL.Path, testRequest.URL.RawQuery, logger)
+
+	assert.Equal(t, expectedTransportName, selectedTransport.Name)
+	assert.Nil(t, err)
+}
+
+func TestShouldFailSelectTransportWhenNoMatches(t *testing.T) {
```
wookie41: `TestShouldFailToSelectTransportWhenNo`

transport/transport.go
```diff
@@ -306,3 +318,69 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+	roundTrippers := make(map[string]http.RoundTripper)
+	transportMatcher := &Matcher{TransportsConfig: clientConf.Transports}
+	maxIdleConnsPerHost := defaultMaxIdleConnsPerHost
+	if len(clientConf.Transports) > 0 {
+		for _, transport := range clientConf.Transports {
+			roundTrippers[transport.Name] = perepareTransport(transport.Properties, maxIdleConnsPerHost)
+		}
+		transportMatcher.RoundTrippers = roundTrippers
+	} else {
+		return transportMatcher, errors.New("Service->Server->Client->Transports config is empty")
```
wookie41: Shouldn't a `nil` be returned on error?

transport/transport.go
```diff
@@ -306,3 +318,69 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+
+// SelectTransportRoundTripper for selecting RoundTripper by request object from transports matcher
+func (m *Matcher) SelectTransportRoundTripper(request *http.Request) (selectedRoundTripper http.RoundTripper, err error) {
+	selectedTransport, err := m.SelectTransportDefinition(request.Method, request.URL.Path, request.URL.RawQuery, log.DefaultLogger)
+	if len(selectedTransport.Name) > 0 {
+		reqID := request.Context().Value(log.ContextreqIDKey)
+		log.Debugf("Request %s - selected transport name: %s (by method: %s, path: %s, queryParams: %s)",
+			reqID, selectedTransport.Name, request.Method, request.URL.Path, request.URL.RawQuery)
+
+		selectedRoundTripper = m.RoundTrippers[selectedTransport.Name]
```
wookie41: this empty line and line number 352 are unnecessary

storages/storages.go
```diff
@@ -28,7 +28,11 @@ func (be *backendError) Err() error {
 }
 
 func (be *backendError) Error() string {
-	return fmt.Sprintf("backend %s responded with error %s", be.backend, be.origErr)
+	if transportDefinitionError, ok := be.origErr.(*transport.DefinitionError); ok {
+		return fmt.Sprintf("backend %s not responded with reason: %s", be.backend, transportDefinitionError.Error())
```
ksitak: Do we need this message at all? Not matched transport has nothing to do with backend itself.

storages/storages.go
```diff
@@ -28,7 +28,11 @@ func (be *backendError) Err() error {
 }
 
 func (be *backendError) Error() string {
```
ksitak: "be" is wrong variable name

config/validator_test.go
```diff
@@ -258,11 +260,201 @@ func TestValidatorShouldFailWithMissingClusterDefinition(t *testing.T) {
...
+	}
+	var size httphandlerconfig.HumanSizeUnits
+	size.SizeInBytes = 2048
+	yamlConfig := PrepareYamlConfig(size, 31, 45, "127.0.0.1:81",
+		"127.0.0.1:1234", "127.0.0.1:1235", nil, transports)
+	valid, _ := yamlConfig.TransportsEntryLogicalValidator()
+	assert.True(t, valid)
+}
+
+func TestValidatorShouldProcessTransportsWithSuccessWithNotDefinedRulesProperties(t *testing.T) {
```
wookie41: `...WhenOnlyOneRuleIsDefined`

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+// compileRule
+func (t *TransportMatcherDefinition) compileRule(regexpRule string) (compiledRule *regexp.Regexp, err error) {
+	if len(regexpRule) > 0 {
+		compiledRule, err = regexp.Compile(regexpRule)
+	}
+	return
+}
+
+// transportFlags for internal matching func
+type transportFlags struct {
```
wookie41: `ruleFlags`

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+// CompiledRules properties
+type CompiledRules struct {
+	MethodRegexp     *regexp.Regexp
+	PathRegexp       *regexp.Regexp
+	QueryParamRegexp *regexp.Regexp
+	IsCompiled       bool
+}
+
+// TransportMatcherDefinition properties
+type TransportMatcherDefinition struct {
```
wookie41: `TransportMatcher` alone sound much better imho

transport/config/config.go
```diff
@@ -0,0 +1,146 @@
...
+		queryParamFlag.matched = transport.CompiledRules.QueryParamRegexp.MatchString(queryParam)
+	} else {
+		queryParamFlag.empty = true
+		queryParamFlag.matched = true
+	}
+	return methodFlag, pathFlag, queryParamFlag
+}
+
+// GetMatchedTransportDefinition returns first details matching with rules from Rules by arguments: method, path, queryParam
+func (t *Transports) GetMatchedTransportDefinition(method, path, queryParam string) (matchedTransport TransportMatcherDefinition, ok bool) {
```
wookie41: `FindTransportForRequestParams`

transport/transport.go
```diff
@@ -306,3 +323,68 @@ func NewMultiTransport(backends []http.RoundTripper,
...
+func (m *Matcher) RoundTrip(request *http.Request) (response *http.Response, err error) {
+	selectedRoundTriper, err := m.SelectTransportRoundTripper(request)
+	if err == nil {
+		return selectedRoundTriper.RoundTrip(request)
+	}
+	return
+}
+
+// SelectTransportRoundTripper for selecting RoundTripper by request object from transports matcher
+func (m *Matcher) SelectTransportRoundTripper(request *http.Request) (selectedRoundTripper http.RoundTripper, err error) {
```
wookie41: `FindTransportForRequest`

transport/config/config.go
bieli: :+1:

README.md
```diff
@@ -240,6 +284,21 @@ Response:
     < Content-Length: 2
     OK
 
+## Transports and Rules with dedicated timeouts
+
+This feature guarantee high availability and better transmission.
```
ksitak: guarantees

README.md
```diff
@@ -240,6 +284,21 @@ Response:
...
+
+For example, when one specific HTTP method has lag we can set timeouts with special 'Rule'.
+Another example, when user adds big chunks by multi upload,
+default timeout needs to be changed with dedicated 'Transport' with 'Rule' for this case.
+
+We have 'Rules' for 'Transports' definitions:
+ - required minimum one item in 'Transports' section
+ - required empty or one property (Method, Path, QueryParam) in 'Rules' section
+ - if empty 'Rules' section is defined it will match with any requests
```
ksitak: if Rules section is empty, the transport will match any requests

PR #78, opened 2018-03-19
Domain style support

wookie41:
Support for domain style support and rewriting domain style to path style.


regions/regions.go
```diff
@@ -44,14 +45,34 @@ func (rg Regions) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	}
 	if rg.defaultRing != nil {
 		return rg.defaultRing.DoRequest(req)
 	}
 	return rg.getNoSuchDomainResponse(req), nil
 }
 
+
+func removeBucketNameFromHost(host string) (string, string){
+	splitted := strings.SplitN(host, ".", 2)
```
mjarco: This trick does not comply with bucket naming specs:
"Bucket names must be a series of one or more labels. Adjacent labels are separated by a single period (.)."
https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

storages/domain_style.go
```diff
@@ -0,0 +1,42 @@
...
+	host := req.Header.Get(HOST)
+	bucket := req.Header.Get(BUCKET)
+
+	if host == "" {
+		return nil, fmt.Errorf("Missing host header, request id %d!", req.Context().Value(log.ContextreqIDKey))
+	}
+
+	if bucket != "" {
+		req.URL.Path = fmt.Sprintf(PATH_STYLE_FORMAT, bucket, req.URL.Path)
+		req.Header.Del(BUCKET)
```
mjarco: Clean headers unconditionally

regions/regions.go
wookie41: Thanks, I didn't know that there can me multiple dots.

wookie41: Done

storages/domain_style.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"net/http"
+	"fmt"
+	"github.com/allegro/akubra/log"
+)
+
+const (
+	HOST              = "X-3yeLjyjQNx"
+	BUCKET            = "X-lejK0EpVZy"
```
mjarco: Why? Do we really want unreadable, cryptic and suspicious-looking internal headers?
BTW. there is no such convention in go that consts are uppercase, follow mixed case for all names.

regions/regions.go
```diff
@@ -44,13 +45,46 @@ func (rg Regions) RoundTrip(req *http.Request) (*http.Response, error) {
...
+	lastSubDomainIndex := strings.LastIndex(originalHost, ".")
+
+	for lastSubDomainIndex != -1 {
+
+		currentHost = originalHost[lastSubDomainIndex+ 1:] + currentHost
+
+		_, ok = rg.multiCluters[currentHost]
+
+		if ok {
+			return currentHost, originalHost[:lastSubDomainIndex], ok
```
mjarco: What if I told you that you may have two different rings one have assigned akubra.net and other other.akubra.net?

wookie41: I'd say "damn, you're right"

storages/domain_style.go
wookie41: @ksitak suggested that we should use a truly random value for the header name. 
We can actually do `X-3yeLjyjQNx-Akubra-Internal-Host` as a compromise

mjarco: X-Akubra-Internal-Host-3yeLjyjQNx will be fair deal

storages/domain_style.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"net/http"
+	"fmt"
+	"github.com/allegro/akubra/log"
+)
+
+const (
+	HOST              = "X-Akubra-Internal-Host-3yeLjyjQNx"
+	BUCKET            = "X-Akubra-Internal-Bucket-lejK0EpVZy"
+	PATH_STYLE_FORMAT = "/%s%s"
```
mjarco: No underscores in go. Mixed caps in all names, even in consts https://golang.org/doc/effective_go.html#mixed-caps

storages/domain_style.go
ksitak: Yeah, I suggested to include some hash in that header to minimize risk that one day we'll decide to use some X-Akubra...(whatever) header in a real world and we forget that we overwrite it inside.

regions/regions.go
wookie41: Fixed

storages/domain_style.go
wookie41: Fixed

config/validator.go
```diff
@@ -142,3 +142,40 @@ func RequestHeaderContentTypeValidator(req http.Request, requiredContentType str
...
 	return 0
 }
+
+
+// DomainsEntryLogicalValidator checks the correctness of "Domains" part of configuration file
+func (c *YamlConfig) DomainsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
+	errList := make([]error, 0)
+	var domains []string
+	for _, regionConf := range c.Regions {
+		for _, domain := range regionConf.Domains {
```
mjarco: regionDomain for clarity

config/validator.go
```diff
@@ -142,3 +142,40 @@ func RequestHeaderContentTypeValidator(req http.Request, requiredContentType str
...
+			for _, alreadyDefinedDomain := range domains {
+				if domain == alreadyDefinedDomain {
+					continue
```
mjarco: this case will occur only if domain was declared more than one, but that should be an error, don't you think?

wookie41: That's actually true

config/validator.go
```diff
@@ -142,3 +142,40 @@ func RequestHeaderContentTypeValidator(req http.Request, requiredContentType str
...
+				}
+				if strings.HasSuffix(domain, "." + alreadyDefinedDomain) {
+					errList = append(
+						errList,
+						fmt.Errorf("Invalid domain %s! There is a domain %s already defined", domain, alreadyDefinedDomain))
+					continue
+				} else if strings.HasSuffix(alreadyDefinedDomain, "." + domain) {
```
mjarco: this if and one above could be merged into single case, with same error "domain %q conflicts with %q"

config/validator.go
wookie41: Done

config/validator.go
wookie41: Done

config/validator.go
```diff
@@ -142,3 +142,40 @@ func RequestHeaderContentTypeValidator(req http.Request, requiredContentType str
...
 }
+
+
+// DomainsEntryLogicalValidator checks the correctness of "Domains" part of configuration file
+func (c *YamlConfig) DomainsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
+	errList := make([]error, 0)
+	var domains []string
+	for _, regionConf := range c.Regions {
+		for _, domain := range regionConf.Domains {
+			for _, alreadyDefinedDomain := range domains {
```
mjarco: You may move body of that for bloc to separate function for clarity (one indentation less)
eg func isAlreadyDefined(current string, oldones []strings)

regions/regions.go
```diff
@@ -44,14 +46,47 @@ func (rg Regions) RoundTrip(req *http.Request) (*http.Response, error) {
...
+
+		if ok {
+			return currentHost, originalHost[:lastSubDomainIndex]
+		}
+
+		currentHost = "." + currentHost
+		originalHost = originalHost[:lastSubDomainIndex]
+		lastSubDomainIndex = strings.LastIndex(originalHost, ".")
+	}
+	return "", ""
```
bieli: `return "", ""` -> `return` - strings are default empty on declartion `func  ... (currentHost, bucket string)`

config/validator.go
```diff
@@ -142,3 +142,41 @@ func RequestHeaderContentTypeValidator(req http.Request, requiredContentType str
...
 }
+
+
+// DomainsEntryLogicalValidator checks the correctness of "Domains" part of configuration file
+func (c *YamlConfig) DomainsEntryLogicalValidator() (valid bool, validationErrors map[string][]error) {
+	errList := make([]error, 0)
+	var domains []string
+	for _, regionConf := range c.Regions {
+		for _, regionDomain := range regionConf.Domains {
+			for _, alreadyDefinedDomain := range domains {
```
mjarco: In case previous comment on that line has missed:
You may move body of that for block to separate function for clarity (one indentation less)
eg func isAlreadyInDefinedDomains(current string, verifiedDomains []string)
(and maybe rename var domain to verifiedDomains)

regions/regions.go
```diff
@@ -44,14 +46,47 @@ func (rg Regions) RoundTrip(req *http.Request) (*http.Response, error) {
...
+		return shardsRing.DoRequest(req)
+
+	}
 	if rg.defaultRing != nil {
 		return rg.defaultRing.DoRequest(req)
 	}
 	return rg.getNoSuchDomainResponse(req), nil
 }
 
+func (rg Regions) findHostInDomainStyle(originalHost string) (currentHost, bucket string) {
```
bieli: Probably Open Source supporting us with subdomain extraction tools -> https://github.com/bobesa/go-domain-util#functions
Why we can't go that way ?

bieli: For me ideal unit test for domain extraction it's based on data with examples and results from fynction -> sth. like this https://github.com/bobesa/go-domain-util/blob/master/domainutil/util_test.go#L53

Of course, when we using our dedicated extractor function.

bieli: If we'll decide using func. `findHostInDomainStyle` I'll be recommending to move func. into `regions/utils` subpackage.

storages/domain_style.go
```diff
@@ -0,0 +1,41 @@
...
+	"net/http"
+	"fmt"
+	"github.com/allegro/akubra/log"
+)
+
+const (
+	// InternalHostHeader is used for rewriting domain style
+	InternalHostHeader   = "X-Akubra-Internal-Host-3yeLjyjQNx"
+	// InternalBucketHeader used for rewriting domain style
+	InternalBucketHeader = "X-Akubra-Internal-Bucket-lejK0EpVZy"
```
bieli: maybe ?!
`X-Custom-Akubra-Internal-Host`
`X-Custom-Akubra-Internal-Bucket`

storages/domain_style.go
```diff
@@ -0,0 +1,41 @@
...
+	// PathStyleFormat is a S3 path style format
+	PathStyleFormat      = "/%s%s"
+)
+
+type domainStyleInterceptor struct {
+	roundTripper http.RoundTripper
```
bieli: `roundTripper` is't necessary in type ... struct, I think

storages/domain_style.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+	if bucket != "" {
+		req.URL.Path = fmt.Sprintf(PathStyleFormat, bucket, req.URL.Path)
+		log.Debugf("Rewritten domain style url to path style url for request ", req.Context().Value(log.ContextreqIDKey))
+	}
+	req.Host = host
+	req.URL.Host = host
+	req.Header.Del(InternalHostHeader)
+	req.Header.Del(InternalBucketHeader)
+	return interceptor.roundTripper.RoundTrip(req)
```
bieli: why not use `interceptor.RoundTrip(req)` instead of `interceptor.roundTripper.RoundTrip(req)` ?

storages/domain_style.go
ksitak: Please, don't waste more time on discussing naming conventions.

config/validator.go
```diff
@@ -142,3 +142,36 @@ func RequestHeaderContentTypeValidator(req http.Request, requiredContentType str
...
+				if regionDomain == alreadyDefinedDomain {
+					errList = append(
+						errList,
+						fmt.Errorf("Invalid domain %s! Domain already defined", regionDomain))
+				} else if strings.HasSuffix(regionDomain, "." + alreadyDefinedDomain) ||
+						  strings.HasSuffix(alreadyDefinedDomain, "." + regionDomain) {
+					errList = append(
+						errList,
+						fmt.Errorf("Invalid domain %s! Domain conflicts with %s", regionDomain, alreadyDefinedDomain))
```
ksitak: Maybe it does not follow some conventions, but what about \n at the end to make it more readable? (and the same with another message above).

storages/domain_style.go
wookie41: `domainStyleInterceptor` is just a wrapper for a `RoundTripper`. It does something with the request (check the `RoundTrip` function below) and delegates to the original `RoundTripper` that you provided in the constructor

storages/domain_style.go
wookie41: Because it would be a recursive call until we run out of memory. Here we delegate to the original `RoundTripper`

sharding/shards_ring.go
```diff
@@ -166,9 +161,9 @@ func (sr ShardsRing) DoRequest(req *http.Request) (resp *http.Response, rerr err
 	if err != nil {
 		return nil, err
 	}
-	isBucketReq := sr.isBucketPath(reqCopy.URL.Path)
+	isBucketReq := utils.IsBucketPath(reqCopy)
```
mjarco: more accurate names: hasBucketPath, isBucketRequest, isBucketOperation

utils/utils.go
```diff
@@ -49,3 +58,22 @@ func ExtractAccessKey(req *http.Request) string {
 	}
 	return strings.TrimSpace(sigChunk[0])
 }
+
+// IsBucketPath tests if a path is a bucket access path
```
mjarco: Misleading comment, it tests whole request as being bucket level access

storages/storages.go
```diff
@@ -70,6 +70,11 @@ type Backend struct {
 	Maintenance bool
 }
 
+// BackendAdapter holds the necessary backend info after decorating the backend
+type BackendAdapter struct {
+	*Backend
+}
```
mjarco: I don' see the point for having this type, all it looks like no-op extension of Backend

PR #79, opened 2018-03-19
More bucket requests handling

mjarco:


storages/merger/versions.go
```diff
@@ -0,0 +1,100 @@
...
+
+func pickVersionResultSet(ps objectsContainer, os objectsContainer, maxKeys int, lbr s3datatypes.ListVersionsResult) s3datatypes.ListVersionsResult {
+	lbr.Version = lbr.Version.FromStringer(ps.first(maxKeys))
+	oLen := maxKeys - len(lbr.Version)
+	lbr.DeleteMarker = lbr.DeleteMarker.FromStringer(os.first(oLen))
+	isTruncated := os.Len()+ps.Len() > maxKeys
+	if !isTruncated {
+		return lbr
+	}
+	// TODO-mj: pack NextContinuatio
```
ksitak: what about it?

storages/response_handler_test.go
```diff
@@ -52,24 +53,57 @@ func responseBuilder(prefixes []CommonPrefix, contents []ObjectInfo, maxKeys int
 	return transport.ResErrTuple{Req: r, Res: resp, Err: nil, Failed: false}, err
 }
 
-func prefixes(prefix ...string) []CommonPrefix {
-	pfs := []CommonPrefix{}
+func responseV2Builder(prefixes []s3datatypes.CommonPrefix, contents []s3datatypes.ObjectInfo, maxKeys int) (transport.ResErrTuple, error) {
+	r, err := http.NewRequest(http.MethodGet, "/bucket", nil)
```
ksitak: r, q... maybe it could be more descriptive?

storages/merger/list.go
```diff
@@ -0,0 +1,134 @@
...
+func MergeListResponses(successes []transport.ResErrTuple) (resp *http.Response, err error) {
+	if len(successes) == 0 {
+		err = fmt.Errorf("No successful responses")
+		return
+	}
+	oContainer := objectsContainer{
+		list: make([]fmt.Stringer, 0),
+		set:  make(map[string]struct{}),
+	}
+	pContainer := objectsContainer{
```
wookie41: what `o` and `p` are? I guess that`o` means `objects`, if so, then maybe a better name would be content container. Same with `p` - `prefixesContainer`?

storages/merger/list.go
```diff
@@ -0,0 +1,134 @@
...
+	"sort"
+	"strconv"
+
+	"github.com/allegro/akubra/log"
+	"github.com/allegro/akubra/storages/merger/s3datatypes"
+	"github.com/allegro/akubra/transport"
+)
+
+// MergeListResponses unifies responses from multiple backends
+func MergeListResponses(successes []transport.ResErrTuple) (resp *http.Response, err error) {
```
wookie41: I would include `Bucket` in method's name, so `MergeListResponses`, because now it suggests, that it can merge any type of responses, that return a list.

storages/auth/s3_signer.go
```diff
@@ -171,6 +171,7 @@ func (srt signAuthServiceRoundTripper) RoundTrip(req *http.Request) (*http.Respo
 	if err != nil {
 		return &http.Response{StatusCode: http.StatusBadRequest, Request: req}, err
 	}
+	log.Println("sign sas ", authHeader)
```
bieli: maybe log.Debug ?

storages/merger/list.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		log.Debugf("ListBucketResult unmarshalling problem %s", err)
+	}
+
+	return lbr
+}
+
+func pickResultSet(keys objectsContainer, prefixes objectsContainer, maxKeys int, listBucketResult s3datatypes.ListBucketResult) s3datatypes.ListBucketResult {
+	listBucketResult.CommonPrefixes = listBucketResult.CommonPrefixes.FromStringer(prefixes.first(maxKeys))
+	keysCount := maxKeys - len(listBucketResult.CommonPrefixes)
```
wookie41: Why `maxKeys - len(listBucketResult.CommonPrefixes)`? Shouldn't it be just `keysCount = len(listBucketResult.CommonPrefixes)`?

storages/merger/list.go
```diff
@@ -0,0 +1,134 @@
...
+	listBucketResult.Contents = listBucketResult.Contents.FromStringer(keys.first(keysCount))
+	isTruncated := keys.Len()+prefixes.Len() > maxKeys
```
wookie41:
```
isNotTruncatedTruncated := keys.Len()+prefixes.Len() < maxKeys
+	if isNotTruncated {
+		return listBucketResult
+	}
```

storages/merger/list.go
```diff
@@ -0,0 +1,134 @@
...
+	bodyBytes := buf.Bytes()
+	err := xml.Unmarshal(bodyBytes, &lbr)
+	if err != nil {
+		log.Debugf("ListBucketResult unmarshalling problem %s", err)
+	}
+
+	return lbr
+}
+
+func pickResultSet(keys objectsContainer, prefixes objectsContainer, maxKeys int, listBucketResult s3datatypes.ListBucketResult) s3datatypes.ListBucketResult {
```
wookie41: `pick` suggest, that this function accepts at least two result sets and pick one of them, but in fact it just (if needed) truncates the response and inserts a marker, so imho a better name would be something like `createResultSet`

storages/merger/listV2.go
```diff
@@ -0,0 +1,125 @@
...
+
+	listBucketV2Result = pickListV2ResultSet(keys, prefixes, maxKeys, listBucketV2Result)
+
+	bodyBytes, err := xml.Marshal(listBucketV2Result)
+	if err != nil {
+		log.Debug("Problem marshalling ObjectStore response body, %s", err)
+		return nil, err
+	}
+	buf := bytes.NewBuffer(bodyBytes)
+	resp.Body = ioutil.NopCloser(buf)
```
wookie41: Shouldn't there be somewhere an initialization of the `resp`? Also, what about other headers that are already in `successes[...].Res`? Shouldn't we merge them somehow?

storages/merger/list.go
```diff
@@ -0,0 +1,134 @@
...
+
+	listBucketResult = pickResultSet(keys, prefixes, maxKeys, listBucketResult)
+
+	bodyBytes, err := xml.Marshal(listBucketResult)
+	if err != nil {
+		log.Debug("Problem marshalling ObjectStore response body, %s", err)
+		return nil, err
+	}
+	buf := bytes.NewBuffer(bodyBytes)
+	resp.Body = ioutil.NopCloser(buf)
```
wookie41: Same question as below, why the `resp` is not initialzied anywhere (does Go automatically initializes the struct with deafault values, when you reference any of its fields?), and what about the headers from the responses in `successes`?

storages/merger/versions.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+
+	}
+	lbr.Version = versions
+	lbr.DeleteMarker = deleteMarkers
+
+	lbr.KeyMarker = lastMarker.GetKey()
+	lbr.VersionIDMarker = lastMarker.GetVersionID()
+
+	isTruncated := keysContainer.Len() > maxKeys
```
wookie41: The logic can be inverted as described in one of the comments above

storages/merger/versions.go
```diff
@@ -0,0 +1,117 @@
...
+	bodyBytes := buf.Bytes()
+	err := xml.Unmarshal(bodyBytes, &lbr)
+	if err != nil {
+		log.Debugf("ListBucketResult unmarshalling problem %s", err)
+	}
+
+	return lbr
+}
+
+func pickVersionResultSet(keysContainer *objectsContainer, maxKeys int, lbr s3datatypes.ListVersionsResult) s3datatypes.ListVersionsResult {
```
wookie41: `pick` -> `create`

storages/merger/listV2.go
```diff
@@ -0,0 +1,125 @@
...
+	bodyBytes := buf.Bytes()
+	err := xml.Unmarshal(bodyBytes, &lbr)
+	if err != nil {
+		log.Debugf("ListBucketResult unmarshalling problem %s", err)
+	}
+
+	return lbr
+}
+
+func pickListV2ResultSet(keys objectsContainer, prefixes objectsContainer, maxKeys int, bucketListV2Result s3datatypes.ListBucketV2Result) s3datatypes.ListBucketV2Result {
```
wookie41: `pick` -> `create`

storages/merger/listV2.go
```diff
@@ -0,0 +1,125 @@
...
+	bucketListV2Result.CommonPrefixes = bucketListV2Result.CommonPrefixes.FromStringer(prefixes.first(maxKeys))
+	keysCount := maxKeys - len(bucketListV2Result.CommonPrefixes)
+	bucketListV2Result.Contents = bucketListV2Result.Contents.FromStringer(keys.first(keysCount))
+	isTruncated := keys.Len()+prefixes.Len() > maxKeys
```
wookie41: The logic can be inverted as described in one of the comments above

storages/response_handler.go
```diff
@@ -92,95 +22,20 @@ func isSuccess(tup transport.ResErrTuple) bool {
...
-
-	buf := &bytes.Buffer{}
-
-	if _, rerr := buf.ReadFrom(resp.Body); rerr != nil {
-		log.Debugf("Problem reading ObjectStore response body, %s", rerr)
-		return lbr
-	}
+func (rm *responseMerger) createResponse(firstTuple transport.ResErrTuple, successes []transport.ResErrTuple) (resp *http.Response, err error) {
+	reqQuery := firstTuple.Req.URL.Query()
+	log.Println("Create response")
```
wookie41: `Debug`?

storages/merger/listV2.go
mjarco: `resp` is one of valid backend responses (see line 32), so headers are appropriate

storages/merger/list.go
mjarco: There is no need for such initialization (see answer above). 
Yes go initialises all values (not just struct types) with type zeros (0, false, "", nil). Struct is 'zero' if all fields are 'zeroes'.
Headers are just fine :)

PR #86, opened 2018-04-23
Added: metrics for CLOSE_WAIT duration, Added: filter synclog read events when in maintenance mode

mjarco:
…ents when maintenance

### DIFF
--- a/httphandler/response_merger.go
+++ b/httphandler/response_merger.go
@@ -3,11 +3,13 @@ package httphandler
 import (
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"time"
 
 	"github.com/allegro/akubra/log"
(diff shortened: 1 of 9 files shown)


httphandler/response_merger.go
```diff
@@ -85,6 +94,16 @@ func (rd *responseMerger) handleFailedResponces(
 	return alreadysent
 }
 
+func shouldBeFilteredInMaintenanceMode(failedTup transport.ResErrTuple) bool {
+	if failedTup.Req.Method == http.MethodPut || failedTup.Req.Method == http.MethodDelete {
```
ksitak: Are we sure that Post is not necessary here?

mjarco: As agreed we'll leave POSTs out for now

PR #89, opened 2018-05-09
sync tasks log by service discovered from consul

bieli:


log/brimapi/logger.go
```diff
@@ -60,3 +73,22 @@ func (lh *LogHook) Fire(entry *logrus.Entry) error {
 	}
 	return discardBody(resp)
 }
+
+// Discovery service URI for service:// protocol
+func (lh *LogHook) discoveryServiceURI(URI string) (*url.URL, error) {
```
wookie41: Method name shouldn't be noun, I can't tell what the method does just by looking at it's signature.

log/brimapi/logger.go
```diff
@@ -60,3 +73,22 @@ func (lh *LogHook) Fire(entry *logrus.Entry) error {
...
+	if !strings.HasPrefix(parsedURI.Scheme, serviceDiscoverySchemeName) {
+		return parsedURI, nil
+	}
+
+	consulClient, err := api.NewClient(api.DefaultConfig())
+	if err != nil {
+		return nil, fmt.Errorf("unable to create Consul client: %s", err)
+	}
+	resolver := service.New(consulClient)
+	return resolver.GetRandomNode(parsedURI.Host)
```
wookie41: This method actually doest a lot of stuff that I didn't expect after just seeing the signature. There is an implicit setup of consul client, it should be more verbose

log/brimapi/logger.go
```diff
@@ -60,3 +73,22 @@ func (lh *LogHook) Fire(entry *logrus.Entry) error {
...
+	parsedURI, err := url.Parse(URI)
+	if err != nil {
+		return nil, err
+	}
+
+	if !strings.HasPrefix(parsedURI.Scheme, serviceDiscoverySchemeName) {
+		return parsedURI, nil
+	}
+
+	consulClient, err := api.NewClient(api.DefaultConfig())
```
wookie41: What are the default timeouts etc?

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+
+// Resolver for discovery service
+type Resolver struct {
+	cache     *cache.Cache
+	client    *api.Client
+	generator *rand.Rand
+}
+
+// GetNodes get service nodes form service discovery
+func (resolver *Resolver) GetNodes(service string) (entries []*api.ServiceEntry) {
```
wookie41: Maybe `FindServiceInstances`?

service/discovery.go
```diff
@@ -0,0 +1,73 @@
+package service
```
wookie41: Why not `discovery`?

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+	"github.com/hashicorp/consul/api"
+	"golang.org/x/sync/syncmap"
+)
+
+const (
+	updateInstancesCheckSeconds = 30
+)
+
+// Services for discovery service
+type Services struct {
```
wookie41: `ServiceDiscoveryClient`

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+	resolver := s.UpdateInstences(serviceName)
+	if resolver != nil {
+		url = resolver.prepareCurrentEndpoint()
+	} else {
+		err = fmt.Errorf("no registered or healtly instances for service: %s", serviceName)
+	}
+	return
+}
+
+// UpdateInstences get service instances form service discovery
```
wookie41: `from`

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
+	value, exists := s.Instances.Load(serviceName)
```
wookie41: What `exists`?

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+		resolver = value.(*Resolver)
+	}
+	if s.Instances == nil || resolver == nil || !exists {
+		s.Instances = new(syncmap.Map)
+		resolver = NewResolver(s.ConsulClient)
+		resolver.updateLastTimestamp()
+	}
+
+	if !exists || time.Now().Unix()-resolver.TTL >= updateInstancesCheckSeconds {
+		if exists {
```
wookie41: Is the `exists` condition really needed here? Because, if exists is `false`, then the condition `if s.Instances == nil || resolver == nil || !exists ` will be true and a new `resolver` will be created.

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+			if !resolver.tryLock() {
+				return resolver
+			}
+		} else {
+			resolver.lock.Lock()
+		}
+		entries := resolver.GetNodesFromConsul(serviceName)
+
+		resolver.endpoints = make([]*url.URL, 0)
```
wookie41: 0 -> `len(entries)`

service/discovery.go
```diff
@@ -0,0 +1,73 @@
...
+	if resolver != nil {
+		url = resolver.prepareCurrentEndpoint()
+	} else {
+		err = fmt.Errorf("no registered or healtly instances for service: %s", serviceName)
+	}
+	return
+}
+
+// UpdateInstences get service instances form service discovery
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
```
wookie41: `GetResolverForService`

service/resolver.go
```diff
@@ -0,0 +1,73 @@
...
+		AllowStale:        true,
+		RequireConsistent: false,
+	})
+	if error == nil {
+		entries = nodes
+	}
+	return
+}
+
+func (r *Resolver) prepareCurrentEndpoint() (currentEndpoint *url.URL) {
```
wookie41: What `current` means in this context? Maybe something like `getHealthyInstanceEndpoint`

service/discovery.go
```diff
@@ -0,0 +1,75 @@
...
+		url = resolver.getHealthyInstanceEndpoint()
+	} else {
+		err = fmt.Errorf("no registered or healtly instances for service: %s", serviceName)
+	}
+	return
+}
+
+// UpdateInstences get service instances from service discovery
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
+	value, instancesExists := s.Instances.Load(serviceName)
```
wookie41: imho `serviceInstancesResolverExists` would be perfect

service/resolver.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/hashicorp/consul/api"
+)
+
+// Resolver for discovery service
+type Resolver struct {
+	consulClient    *api.Client
+	endpoints       []*url.URL
+	currentEndpoint *url.URL
+	generator       *rand.Rand
+	TTL             int64
```
bnogas: Maybe lastUpdateTimestamp ?

service/discovery.go
```diff
@@ -0,0 +1,75 @@
...
+// DefaultCacheInvalidationTimeout for updating instances
+const (
+	DefaultCacheInvalidationTimeout = 30
+)
+
+// Services for discovery service
+type Services struct {
+	ConsulClient             *api.Client
+	Instances                *syncmap.Map
+	CacheInvalidationTimeout int64
```
bnogas: I'd name it CacheTTL

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+	"strings"
+	"time"
+
+	"github.com/allegro/akubra/service"
+	"github.com/hashicorp/consul/api"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	uploadSynctasksURI         = "/v1/processes/uploadsynctasks"
```
wookie41: Maybe we can extract this to the config?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+	httpRequestTimeout         = time.Second * 3
+	httpConsulRequestTimeout   = time.Second * 2
+)
+
+var (
+	discoveryServices *service.Services
+	httpClient        *http.Client
+)
+
+func init() {
```
wookie41: what is being initialized here?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+	httpClient = &http.Client{
```
wookie41: Can't we use the client that we already have configured?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+		Timeout: httpRequestTimeout,
+	}
+	consulConfig := api.DefaultConfig()
+	api.DefaultConfig().Transport = &http.Transport{
```
wookie41: Shouldn't this be `consulConfig.Transport = ...`?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
+package brimapi
```
wookie41: Consul doens't have anything to do with `brim`, it should be in seperate package

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+	api.DefaultConfig().Transport = &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   httpConsulRequestTimeout,
+			DualStack: true,
+		}).DialContext,
+	}
+	consulClient, err := api.NewClient(consulConfig)
+	if err != nil {
+		fmt.Printf("unable to create Consul client: %s", err)
```
wookie41: What happens when the discovery service (consul in this case) couldn't be configured?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func doRequestWithDiscoveryService(lh *LogHook, httpClient *http.Client, payload string) (resp *http.Response, endpoint string, err error) {
+	uri, discoveryServiceErr := lh.getServiceURIFromDiscoveryService(lh.Host)
+	if discoveryServiceErr != nil {
+		return nil, "", discoveryServiceErr
+	}
+	uri.Path = uploadSynctasksURI
+	endpoint = uri.String()
+	logrus.Debugf("sync task endpoint: %s", endpoint)
```
wookie41: `logrus.Debugf`?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+func doRequest(lh *LogHook, httpClient *http.Client, payload string) (endpoint string, err error) {
+	resp, endpoint, err := doRequestWithDiscoveryService(lh, httpClient, payload)
+	if err != nil {
+		return
+	}
+	return endpoint, discardBody(resp)
+}
+
+func doRequestWithDiscoveryService(lh *LogHook, httpClient *http.Client, payload string) (resp *http.Response, endpoint string, err error) {
+	uri, discoveryServiceErr := lh.getServiceURIFromDiscoveryService(lh.Host)
```
wookie41: `uri` -> `serviceInstanceAddress`

service/discovery.go
```diff
@@ -0,0 +1,75 @@
...
+	return
+}
+
+// UpdateInstences get service instances from service discovery
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
+	value, instancesExists := s.Instances.Load(serviceName)
+	if instancesExists {
+		resolver = value.(*Resolver)
+	} else {
+		s.Instances = new(syncmap.Map)
```
wookie41: Wait, if I understand correctly, then assuming that `s.Instances` has entries for services `B` and `C` and a resolver for service `A` doesn't exist, then we flush the entries for services `B` and `C`.  
Right?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	consulClient, err := api.NewClient(consulConfig)
+	if err != nil {
+		fmt.Printf("unable to create Consul client: %s", err)
+	}
+
+	discoveryServices = service.New(consulClient, service.DefaultCacheInvalidationTimeout)
+}
+
+func doRequest(lh *LogHook, httpClient *http.Client, payload string) (endpoint string, err error) {
```
wookie41: Why these methods (this and the one below)  accept a `payload`? What if I wanted to send a stream of bytes?

log/brimapi/client.go
```diff
@@ -0,0 +1,103 @@
...
+		bytes.NewBuffer([]byte(payload)))
+	if requestErr != nil {
+		return nil, endpoint, requestErr
+	}
+	req.SetBasicAuth(lh.Creds.User, lh.Creds.Pass)
+	resp, err = httpClient.Do(req)
+	return
+}
+
+func (lh *LogHook) getServiceURIFromDiscoveryService(URI string) (*url.URL, error) {
```
wookie41: This should be in `logger.go`

service/discovery.go
```diff
@@ -0,0 +1,75 @@
...
+	if resolver != nil && len(resolver.endpoints) > 0 {
+		url = resolver.getHealthyInstanceEndpoint()
+	} else {
+		err = fmt.Errorf("no registered or healtly instances for service: %s", serviceName)
+	}
+	return
+}
+
+// UpdateInstences get service instances from service discovery
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
```
wookie41: Shouldn't this method be called `GetServiceInstancesResolver`?

log/brimapi/client.go
bieli: We have dedicated file `client.go` for all things related with discovery service and `logger.go` is clean

log/brimapi/client.go
bieli: yes

log/brimapi/client.go
bieli: `discoveryServices` instance with related `httpClient` + `consulConfig`

service/discovery.go
```diff
@@ -0,0 +1,75 @@
...
+	if instancesExists {
+		resolver = value.(*Resolver)
+	} else {
+		s.Instances = new(syncmap.Map)
+		resolver = NewResolver(s.ConsulClient)
+		resolver.updateLastTimestamp()
+	}
+
+	if !instancesExists || time.Now().Unix()-resolver.TTL >= s.CacheInvalidationTimeout {
+		if instancesExists {
```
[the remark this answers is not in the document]
bieli: I can't reproducing this problem in practice. I do test with 100 goroutines and 50 service instances in consul with updating cache/instances time in two cases: 1, 10 seconds. Test time ~4h and I can't see cases with updating cache twice.

discovery/service.go
```diff
@@ -0,0 +1,75 @@
...
+	if resolver != nil && len(resolver.endpoints) > 0 {
+		url = resolver.getHealthyInstanceEndpoint()
+	} else {
+		err = fmt.Errorf("no registered or healtly instances for service: %s", serviceName)
+	}
+	return
+}
+
+// UpdateInstences get service instances from service discovery
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
```
ksitak: instances

service/discovery.go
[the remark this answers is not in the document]
ksitak: Łukasz is right. I accept the fact it may happen on Akubra startup, when cache is empty, but double (or, in fact, triple...) update should never happen when cache is already filled with instance data.

discovery/service.go
```diff
@@ -0,0 +1,75 @@
...
+		resolver.updateLastTimestamp()
+		s.Instances.Store(serviceName, resolver)
+		resolver.lock.Unlock()
+	}
+
+	return resolver
+}
+
+// New Services constructor
+func New(consulClient *api.Client, cacheInvalidationTimeout int64) *Services {
```
ksitak: New what?

discovery/resolver.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/hashicorp/consul/api"
+)
+
+// Resolver for discovery service
+type Resolver struct {
+	consulClient    *api.Client
+	endpoints       []*url.URL
+	currentEndpoint *url.URL
+	generator       *rand.Rand
+	CacheTTL        int64
```
ksitak: lastUpdateTime?

discovery/service.go
```diff
@@ -0,0 +1,75 @@
...
+
+// UpdateInstences get service instances from service discovery
+func (s *Services) UpdateInstences(serviceName string) (resolver *Resolver) {
+	value, instancesExists := s.Instances.Load(serviceName)
+	if instancesExists {
+		resolver = value.(*Resolver)
+	} else {
+		s.Instances = new(syncmap.Map)
+		resolver = NewResolver(s.ConsulClient)
+		resolver.updateLastTimestamp()
```
ksitak: we set it below, is it really needed here?

discovery/service.go
```diff
@@ -0,0 +1,75 @@
...
+	if !instancesExists || time.Now().Unix()-resolver.CacheTTL >= s.CacheInvalidationTimeout {
+		if instancesExists {
+			if !resolver.tryLock() {
+				return resolver
+			}
+		} else {
+			resolver.lock.Lock()
+		}
+		entries := resolver.GetNodesFromConsul(serviceName)
+
```
ksitak: what if resolver fails and entries is empty? what's the point of replacing existing cache with empty value?

discovery/service.go
```diff
@@ -0,0 +1,75 @@
...
+		resolver.endpoints = make([]*url.URL, 0)
+		for _, entry := range entries {
+			resolver.endpoints = append(resolver.endpoints, serviceEntryToURL(entry))
+		}
+		resolver.updateLastTimestamp()
```
ksitak: All this code seems quite weird for me. We get a value from a resolver to put a value in a resolver using a method from a resolver. Why all this can't happen in the resolver itself?

discovery/service_mock_test.go
```diff
@@ -0,0 +1,117 @@
...
+	consulClientMock := prepareConsulClientMock(host, port, 1)
+	services := NewServices(consulClientMock, 1)
+
+	instance, err := services.GetEndpoint("my-service")
+
+	assert.NoError(t, err)
+	assert.Equal(t, instance.Host, expectedHost)
+}
+
+func TestShouldGetEndpointWhenTwoInstancesExists(t *testing.T) {
```
ksitak: exist

discovery/service_mock_test.go
```diff
@@ -0,0 +1,117 @@
...
+	consulClientMock := prepareConsulClientMock(host, port, 1)
+	services := NewServices(consulClientMock, 1)
+
+	instance, err := services.GetEndpoint("my-service")
+
+	require.NoError(t, err)
+	require.Equal(t, instance.Host, expectedHost)
+}
+
+func TestShouldGetEndpointWhenTwoInstancesExists(t *testing.T) {
```
ksitak: exist

discovery/service_mock_test.go
```diff
@@ -0,0 +1,117 @@
...
+	consulClientMock := prepareConsulClientMock(host, port, 2)
+	services := NewServices(consulClientMock, 5)
+
+	instance, err := services.GetEndpoint("my-service")
+
+	require.NoError(t, err)
+	require.True(t, instance.Host == expectedHost1 || instance.Host == expectedHost2)
+}
+
+func TestShouldGetEndpointReturnEndpointAfterTwiceCalling(t *testing.T) {
```
ksitak: If this tests checks double calling then what is the sleep inside for?

bieli: simulating delay for next cache revalidation

PR #92, opened 2018-06-21
Mem leak2

mjarco:


storages/replicator.go
```diff
@@ -28,8 +29,14 @@ func newReplicationClient(backends []*backend.Backend) client {
 func (rc *ReplicationClient) Do(request *http.Request) <-chan BackendResponse {
 	responsesChan := make(chan BackendResponse)
 	wg := sync.WaitGroup{}
-	ctx, cancelFunc := context.WithCancel(request.Context())
+	reqIDValue, ok := request.Context().Value(log.ContextreqIDKey).(string)
+	if !ok {
+		reqIDValue = ""
```
wookie41: Why can't we generate an id if it's not in the context?

mjarco: To distinguish between lost request ID and legit request ID

storages/backend/backend.go
```diff
@@ -60,24 +62,44 @@ func (b *Backend) collectMetrics(resp *http.Response, err error, since time.Time
...
+		backendName = br.Backend.Name
+	}
 	if br.Response == nil || br.Response.Body == nil {
+		log.Debugf("ResponseBody for request %s is nil so cannot be closed - backend: %s", br.ReqID(), backendName)
 		return nil
 	}
 	_, err := io.Copy(ioutil.Discard, br.Response.Body)
 	if err != nil {
 		log.Printf("Discard body error %s", err)
 		return err
```
wookie41: Don't we have to close the body here as well? Or at least try to close it? (if the drain fails)

storages/multipart_round_tripper.go
```diff
@@ -58,7 +58,7 @@ func (multiPartRoundTripper *MultiPartRoundTripper) Do(request *http.Request) <-
 	if !multiPartRoundTripper.canHandleMultiUpload() {
 		log.Debugf("Multi upload for %s failed - no backends available.", request.URL.Path)
 		go func() {
-			backendResponseChannel <- BackendResponse{Response: nil, Error: errors.New("Can't handle multi upload")}
+			backendResponseChannel <- BackendResponse{Request: request, Response: nil, Error: errors.New("Can't handle multi upload")}
```
wookie41: Maybe extract `errors.New("Can't handle multi upload")` to some global, so we can later compare it in conditions?

storages/multipart_round_tripper.go
```diff
@@ -69,7 +69,7 @@ func (multiPartRoundTripper *MultiPartRoundTripper) Do(request *http.Request) <-
 	if backendSelectError != nil {
 		log.Debugf("Multi upload failed for %s - %s", backendSelectError, request.URL.Path)
 		go func() {
-			backendResponseChannel <- BackendResponse{Response: nil, Error: errors.New("Can't handle multi upload")}
+			backendResponseChannel <- BackendResponse{Request: request, Response: nil, Error: errors.New("Can't handle multi upload")}
```
wookie41: Don't we want to distinguish situations when can't simply handle a multipart upload and when the multipart upload failed? (we return the same error now - `errors.New("Can't handle multi upload")`)

PR #95, opened 2018-07-19
Fixup some English in README

deejay1:


README.md
```diff
@@ -12,27 +12,27 @@
...
 
 Akubra way is to put files in all storages at once by copying requests to multiple
-backends. I case one of clusters rejects request it logs that event, and syncronizes
+backends. I case one if clusters rejects request it logs that event, and synchronizes
 troublesome object with an independent process.
 
 ### Seamless storage space extension with new storage clusters
-Akubra has sharding capabilities. You may easily configure new backends with
+Akubra has sharding capabilities. You can easily configure new backends with
 weigths and append them to regions cluster pool.
```
ch40s3k: s/weigths/weights

README.md
```diff
@@ -12,27 +12,27 @@
...
-at once. That's what Akubra does, with a minimum memory and cpu footprint.
+Keeping synchronized storage clusters, which handle great volume of new objects
+(about 300k obj/h), is most efficient by feeding them with all incoming data
+at once. That's what Akubra does, with a minimal memory and cpu footprint.
 
-Synchronizing S3 storages offline is almost impossible with a high volume traffic.
+Synchronizing S3 storages offline is almost impossible with a high volume of traffic.
 It would require keeping track of new objects (or periodical bucket listing),
-downloading and uploading them to other storage. It's slow, expensive and hard
+downloading and uploading them to the other storage. It's slow, expensive and hard
```
ksitak: Not sure if "the" is necessary here.

README.md
```diff
@@ -12,27 +12,27 @@
...
 to implement robustly.
 
 Akubra way is to put files in all storages at once by copying requests to multiple
-backends. I case one of clusters rejects request it logs that event, and syncronizes
+backends. I case one if clusters rejects request it logs that event, and synchronizes
```
ksitak: If one of the clusters rejects...

README.md
deejay1: I think it's ok here, because we point at a particular storage, if we would put here `other storages` we would omit the `the`.

README.md
deejay1: Good catch, sad that it was after the merge :)

PR #98, opened 2018-07-24
Instead of http status 203 append header

mjarco:


storages/merger/bucket_meta_data.go
```diff
@@ -17,17 +22,22 @@ func MergePartially(firstResponse backend.Response, successes []backend.Response
 	if err != nil {
 		log.Printf("Could not read first response body reqID %s, reason %s", firstResponse.ReqID(), err)
 	}
-	firstResponse.DiscardBody()
-	statusCode := http.StatusOK
+	err = firstResponse.DiscardBody()
+	if err != nil {
+		log.Printf("Cannot close first response body reqId %s, reason: %s", firstResponse.ReqID(), err)
```
wookie41: Doest `DiscardBody` closes the body?

storages/merger/bucket_meta_data.go
```diff
@@ -17,17 +22,22 @@ func MergePartially(firstResponse backend.Response, successes []backend.Response
 	if err != nil {
 		log.Printf("Could not read first response body reqID %s, reason %s", firstResponse.ReqID(), err)
```
wookie41: Shouldn't there be a `return`?

storages/merger/bucket_meta_data.go
wookie41: `Cannot close` -> `Cannot discard`
And does `DiscardBody` closes the `body` in case it can't be discarded?

mjarco: Tries to, if ends up with error returns it

storages/merger/bucket_meta_data.go
mjarco: No, we want to return any response we have

PR #102, opened 2018-09-18
Fixed: invalid read od closed body error

mjarco:


sharding/shards_ring.go
```diff
@@ -85,17 +85,24 @@ func copyRequest(origReq *http.Request) (*http.Request, error) {
...
 		}
 	}
+
 	if origReq.Body != nil {
 		buf := &bytes.Buffer{}
 		n, err := io.Copy(buf, origReq.Body)
 		if err != nil {
 			return nil, err
 		}
+		err = origReq.Body.Close()
```
wookie41: Shouldn't this albo be inside the `if` above?

PR #107, opened 2018-10-12
Balancing

mjarco:


balancing/balance_breaker.go
```diff
@@ -0,0 +1,531 @@
...
+	"github.com/allegro/akubra/storages/backend"
+	"github.com/allegro/akubra/storages/config"
+)
+
+// ResponseTimeBalancer proxies calls to balancing nodes
+type ResponseTimeBalancer struct {
+	Nodes []Node
+}
+
+// Elect elects node and call it with args
```
wookie41: `calls`

balancing/balance_breaker.go
```diff
@@ -0,0 +1,541 @@
...
+func newLenLimitCounter(retention int) *lenLimitCounter {
+	return &lenLimitCounter{
+		values: make([]float64, retention, retention),
+	}
+}
+
+type lenLimitCounter struct {
+	values  []float64
+	nextIdx int
+	mx      sync.Mutex
```
wookie41: the structure has a mutex as a member, but none of it's metods actually use it - `Add` etc.

balancing/balance_breaker.go
```diff
@@ -0,0 +1,541 @@
...
+		return breaker.checkStateHalfOpen(exceeded)
+	}
+
+	if exceeded {
+		breaker.openBreaker()
+	}
+	return exceeded
+}
+
+func (breaker *NodeBreaker) checkStateHalfOpen(exceeded bool) bool {
```
wookie41: `isHalfOpen`

balancing/balance_breaker.go
```diff
@@ -0,0 +1,531 @@
...
+		if elected == nil {
+			elected = node
+		}
+
+		if nodeWeight(node) < nodeWeight(elected) {
+			elected = node
+		}
+	}
+	if elected == nil {
+		return nil, ErrNoActiveNodes
```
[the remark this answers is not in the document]
mjarco: Caller will report such event

balancing/balance_breaker.go
```diff
@@ -0,0 +1,531 @@
...
+		now:        time.Now,
+	}
+}
+
+// CallMeter implements Node interface
+type CallMeter struct {
+	retention  time.Duration
+	resolution time.Duration
+	calls      int
+	now        func() time.Time
```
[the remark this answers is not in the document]
mjarco: What you'd think if `now` would be a method of any object? What's your naming proposal?

balancing/balance_breaker.go
```diff
@@ -0,0 +1,531 @@
...
+func (meter *CallMeter) Update(duration time.Duration) {
+	series := meter.histogram.pickSeries(meter.now())
+	if series == nil {
+		return
+	}
+	series.Add(float64(duration), meter.now())
+}
+
+// Calls returns number of calls in last bucket
+func (meter *CallMeter) Calls() float64 {
```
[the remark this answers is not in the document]
mjarco: It's interface implementation. `CallsInLastBucket` looks weird in Node interface definition

balancing/balance_breaker.go
```diff
@@ -0,0 +1,541 @@
...
+		mx:         sync.Mutex{},
+	}
+}
+
+type histogram struct {
+	t0         time.Time
+	retention  time.Duration
+	resolution time.Duration
+	data       []*dataSeries
+	now        func() time.Time
```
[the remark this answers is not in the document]
mjarco: "Why would you have the type in variable name?" ;)
Is now() something different than currentTime()?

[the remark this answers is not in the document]
wookie41: :D

It's not, but I found `currentTimeFunc` more descriptive

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+// SetActive sets meter state
+func (meter *CallMeter) SetActive(active bool) {
+	if meter.IsActive() && !active {
+		meter.inActiveSince = meter.now()
+	}
+	if !meter.IsActive() && active {
+		meter.histogram.shiftData(meter.now().Sub(meter.inActiveSince))
+		meter.inActiveSince = time.Time{}
```
wookie41: Maybe `nil` instead of`time.Time`, when switching from inactive to active?

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+	timeLimitPercentile float64
+	timeData            *lengthDelimitedCounter
+	successData         *lengthDelimitedCounter
+	now                 func() time.Time
+	closeDelay          time.Duration
+	maxDelay            time.Duration
+	state               *openStateTracker
+}
+
+// Record collects call data and returns bool if breaker should be open
```
wookie41: `open` -> `opened`

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+	breaker.timeData.Add(float64(duration))
+	successValue := float64(1)
+	if success {
+		successValue = float64(0)
+	}
+	breaker.successData.Add(successValue)
+	return breaker.ShouldOpen()
+}
+
+// ShouldOpen checks if breaker should be open
```
wookie41: as above

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+type openStateTracker struct {
+	state          breakerState
+	lastChange     time.Time
+	changeDelay    time.Duration
+	maxDelay       time.Duration
+	closeIteration float64
+}
+
+func (tracker *openStateTracker) currentDelay() time.Duration {
+	multiplier := int(math.Pow(2, tracker.closeIteration))
```
wookie41: `closeIteration` is never decremented/reset, even if a backend is healthy for a long time. 
An interesting case popped up in my head.
> initially the backend is not doing so well nad its `closeIteration` gets incremented until it reaches some high value
> the backend becomes healthy and stays healthy for a long time, but it's `closeIteration` is already big
> the backend misbehaves for a brief second and gets taken out of the pool of healthy nodes for a long time, due to it's `closeIteration`
> we wait a long time until we plug the backend back in, even though it' already healthy and might provide us with the best timings

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+func (h *histogram) index(now time.Time) int {
+	sinceStart := float64(now.Sub(h.t0))
+	idx := math.Floor(sinceStart / float64(h.resolution))
+	return int(idx)
+}
+
+func (h *histogram) cellsCount() int {
+	return int(math.Ceil(float64(h.retention)/float64(h.resolution))) + 1
+}
+
```
wookie41: Some of the methods of `histogram` are protected by a mutex, but some are not. `shiftData` mutates the state heavily, yet it doesn't lock the data it mutates. If (or when)`shiftData` and `unshiftData` run concurrently bad things may happen.

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+func (series *dataSeries) Add(value float64, dateTime time.Time) {
+	series.mx.Lock()
+	defer series.mx.Unlock()
+	series.data = append(series.data, &timeValue{dateTime, value})
+}
+
+func (series *dataSeries) ValueRangeFun(timeStart, timeEnd time.Time, fun func(*timeValue)) []float64 {
+	dataRange := []float64{}
```
wookie41: `dataRange` is unused

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+		meter.inActiveSince = meter.now()
+	}
+	if !meter.IsActive() && active {
+		meter.histogram.shiftData(meter.now().Sub(meter.inActiveSince))
+		meter.inActiveSince = time.Time{}
+	}
+}
+
+// TimeSpent returns float64 repesentation of time spent in execution
+func (meter *CallMeter) TimeSpent() float64 {
```
wookie41: Won't this be too expensive to calculate It happens every time we want to know the node's weight, which is basically every call. If there are a lot rps, then each call we have to traverse each series in a histogram for all of the active nodes to elect a node. 
We could simply aggregate this instead of calculating.

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+}
+
+// IsActive checks Breaker status propagates it to Node compound
+func (ms *MeasuredStorage) IsActive() bool {
+	isActive := !ms.Breaker.ShouldOpen()
+	ms.Node.SetActive(isActive)
+	return ms.Node.IsActive()
+}
+
+func raportMetrics(rt http.RoundTripper, since time.Time, open bool) {
```
wookie41: report

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+	log.Debugf("MeasuredStorage %s: Request %s took %s was successful: %t, opened breaker %t\n", ms.Name, reqID, duration, success, open)
+
+	ms.Node.UpdateTimeSpent(duration)
+	ms.Node.SetActive(!open)
+	raportMetrics(ms.RoundTripper, start, open)
+	return resp, err
+}
+
+// IsActive checks Breaker status propagates it to Node compound
+func (ms *MeasuredStorage) IsActive() bool {
```
wookie41: i don't like the fact that the method's name hides what it's actually doing and you have to take a look at the doc to know the whole truth.

httphandler/log.go
```diff
@@ -16,7 +16,7 @@ type AccessMessageData struct {
 	Path       string  `json:"path"`
 	UserAgent  string  `json:"useragent"`
 	StatusCode int     `json:"status"`
-	Duration   float64 `json:"duration"`
+	Duration   float64 `json:"duration_ms"`
```
wookie41: isn't json camel case? (duration_ms)

httphandler/roundtripper_decorators.go
```diff
@@ -25,7 +25,7 @@ func (lrt *loggingRoundTripper) RoundTrip(req *http.Request) (resp *http.Respons
 	timeStart := time.Now()
 	resp, err = lrt.roundTripper.RoundTrip(req)
 
-	duration := time.Since(timeStart).Seconds()
+	duration := time.Since(timeStart).Seconds() * 1000
```
wookie41: why multiply by 1000?

storages/shardclient.go
```diff
@@ -0,0 +1,89 @@
...
+		log.Debugf("Request %s, processed by balancer error %s", reqID, err)
+		return resp, err
+
+	}
+	log.Debug("Request %s processed by dispatcher, reqId")
+	return c.requestDispatcher.Dispatch(req)
+}
+
+func (c *ShardClient) balancerRoundTrip(req *http.Request) (resp *http.Response, err error) {
+	notFoundNodes := []balancing.Node{}
```
wookie41: maybe its worth it to add entries to sync log for each of the nodes in `notFoundNodes` after we've found a node that was able to process the request?

balancing/balance_breaker.go
```diff
@@ -0,0 +1,615 @@
...
+}
+func (h *histogram) shiftData(delta time.Duration) {
+	newT0 := h.t0.Add(delta)
+	if newT0.After(h.now()) {
+		return
+	}
+	h.t0 = newT0
+	for _, series := range h.data {
+		for _, value := range series.data {
+			value.date = value.date.Add(delta)
```
wookie41: This loop seems costly. We could instead accumulate the delta inside `histogram` and later, when someone calls `ValueRangeFun` just apply to delta to `timeStart` and `timeEnd` once at the beginning of the method

balancing/balance_breaker.go
```diff
@@ -522,24 +521,28 @@ func (ms *MeasuredStorage) RoundTrip(req *http.Request) (*http.Response, error)
 	log.Debugf("MeasuredStorage %s: Got request id %s\n", ms.Name, reqID)
 	resp, err := ms.RoundTripper.RoundTrip(req)
 	duration := time.Since(start)
-	success := backend.IsSuccessful(resp, err)
+	success := backendSuccess(resp, err)
 	open := ms.Breaker.Record(duration, success)
-	log.Debugf("MeasuredStorage %s: Request %s took %s was successful: %t, opened breaker %t\n", ms.Name, reqID, duration, success, open)
+	log.Debugf("s %s: Request %s took %s was successful: %t, opened breaker %t\n", ms.Name, reqID, duration, success, open)
```
wookie41: You replaced `MeasuredStorage` with `s` in the log message. Is this intentional?

PR #112, opened 2018-11-29
Watchdog

wookie41:


db-migrations/migration.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE consistency_record (
+  request_id      CHARACTER(36) PRIMARY KEY ,
+  object_id       CHARACTER VARYING(1024) NOT NULL,
+  method          CHARACTER VARYING(8) NOT NULL,
+  cluster_name    CHARACTER VARYING(64) NOT NULL,
+  access_key      CHARACTER VARYING(128) NOT NULL,
+  execution_date  TIMESTAMP NOT NULL,
```
bieli: Maybe `TIMESTAMP NOT NULL DEFAULT NOW()` it will be good point ? As I remember we talk about it.

PR #113, opened 2018-11-30
Watchdog

wookie41:
ConsistencyWatchod uses a 'write-ahead'-style log which allows akubra to keep consistency among storages.


storages/request_dispatcher.go
```diff
@@ -14,60 +21,116 @@ type dispatcher interface {
...
+	isMultiPartRequest := isMultiPartUploadRequest(request)
+	isInitiateMultiPart := isInitiateMultiPartUploadRequest(request)
+
+	consistencyCondition := rd.watchdog != nil && len(rd.Backends) > 1
+
+	methodCondition := (http.MethodPut == request.Method && !isMultiPartRequest) ||
+						http.MethodDelete == request.Method ||
+						(http.MethodPost == request.Method && isInitiateMultiPart)
+
+	pathCondition := !utils.IsBucketPath(request.URL.Path)
```
mjarco: What if we would skip this condition? Maybe watchdog & brim could fix client bucket listings on storage failure?

storages/response_picker.go
```diff
@@ -164,8 +170,37 @@ func (drp *deleteResponsePicker) pullResponses(out chan<- BackendResponse) {
 	close(drp.syncLogReady)
 }
 
+
+func pullResponsesWatchdog(drp *baseDeleteResponsePicker, out chan<- BackendResponse) {
+	shouldSend := false
+	emptyResponse := BackendResponse{}
```
mjarco: Move it to package variable, there is no need for creating that on each request. GC will be grateful

storages/response_picker.go
```diff
@@ -164,8 +170,37 @@ func (drp *deleteResponsePicker) pullResponses(out chan<- BackendResponse) {
 	close(drp.syncLogReady)
 }
 
+
+func pullResponsesWatchdog(drp *baseDeleteResponsePicker, out chan<- BackendResponse) {
```
mjarco: Why we have different behaviour for watched requests?

storages/request_dispatcher.go
```diff
@@ -14,60 +21,116 @@ type dispatcher interface {
...
+		watchdog:                  watchdog,
+		watchdogRecordFactory:     watchdogRecordFactory,
 	}
 }
 
 // Dispatch creates and calls replicators and response pickers
 func (rd *RequestDispatcher) Dispatch(request *http.Request) (*http.Response, error) {
+	storageRequest := &Request{request, nil, nil}
+	if rd.shouldUseConsistencyWatchdog(request) {
+		recordedRequest, err := rd.createAndInsertRecordFor(request)
```
mjarco: I think this would be clearer:
var err error
storageRequest, err := rd.createAndInsertRecordFor(request)
than setting properties one by one

storages/request_dispatcher.go
```diff
@@ -14,60 +21,116 @@ type dispatcher interface {
...
 		pickResponsePickerFactory: defaultResponsePickerFactory,
 		pickClientFactory:         defaultReplicationClientFactory,
+		watchdog:                  watchdog,
+		watchdogRecordFactory:     watchdogRecordFactory,
 	}
 }
 
 // Dispatch creates and calls replicators and response pickers
 func (rd *RequestDispatcher) Dispatch(request *http.Request) (*http.Response, error) {
+	storageRequest := &Request{request, nil, nil}
```
mjarco: Don't pass `zero values` in initialisation they are by deafult. BTW name each field 
&Request{request, nil, nil} -> &Request{Request: request}

storages/request_dispatcher.go
```diff
@@ -14,60 +21,116 @@ type dispatcher interface {
...
+		record.ExecutionDate = record.ExecutionDate.Add(oneWeek).Add(time.Minute * 10)
+	}
+	deleteMarker, err := rd.watchdog.Insert(record)
+	if err != nil {
+		return nil, err
+	}
+	return &Request{
+		request,
+		record,
+		deleteMarker,
```
mjarco: Please initialise with fields names 
&Request{
		Request: request,
		record: record,
		marker: deleteMarker,
}

storages/response_picker.go
wookie41: Because the current picker requires a DELETE request to be successful on all of the backed.

storages/request_dispatcher.go
wookie41: I don't understand why would we want  to replace `err := ...` with `var err error`. Can you explain?

mjarco: sry, my fault there is no need to declare err separately

mjarco: goal was to eliminate recordedRequest variable

PR #114, opened 2018-12-21
Signals handling SIGHUP triggers configuration reload, SIGINT gracefu…

mjarco:
…lly shutsdown server

### DIFF
--- a/main.go
+++ b/main.go
@@ -51,7 +51,7 @@ func main() {
 	kingpin.Parse()
 	conf, err := parseConfig(*configFile)
 	if err != nil {
-		log.Fatalf("Configuration corrupted: %s")
+		log.Fatalf("Configuration corrupted: %s", err)
 	}
 
 	if *testConfig {
@@ -124,7 +124,7 @@ func (s *service) start() (err error) {
 	s.ctx = context.Background()
 	s.handler, err = s.createHandler(s.config)
 	if err != nil {
-		log.Fatal("Handler creation error: %s", err)
+		log.Fatalf("Handler creation error: %s", err)
 	}
 	srv := &http.Server{
 		Addr:         s.config.Service.Server.Listen,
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -52,41 +49,45 @@ func main() {
...
-		log.Fatalf("Improperly configured %s", err)
+		log.Fatalf("Configuration corrupted: %s", err)
 	}
 
-	valid, errs := config.ValidateConf(conf.YamlConfig, true)
-	if !valid {
-		log.Printf("YAML validation - errors: %q", errs)
-		os.Exit(1)
+	if *testConfig {
+		os.Exit(0)
```
wookie41: I don't understand. Why we exit with success when config is defined?

mjarco: It's configuration dry-run ./akubra -c config.yaml -t. Like nginx -t validates changed configs

main.go
```diff
@@ -105,72 +106,125 @@ func mkServiceLogs(logConf logconfig.LoggingConfig) (syncLog, clusterSyncLog, ac
...
-			Handler:      handler,
-			ReadTimeout:  s.config.Service.Server.ReadTimeout.Duration,
-			WriteTimeout: s.config.Service.Server.WriteTimeout.Duration,
-		},
-		Timeout: s.config.Service.Server.ShutdownTimeout.Duration,
-	}
-
-	srv.SetKeepAlivesEnabled(true)
-	listener, err := net.Listen("tcp", s.config.Service.Server.Listen)
+	err = metrics.Init(conf.Metrics)
```
bieli: My proposition it's moving `err = metrics.Init(conf.Metrics)` line outside HUP signal block - I suspect many goroutines on stack after recive many HUP signals.

main.go
```diff
@@ -180,19 +234,13 @@ func (s *service) startTechnicalEndpoint() {
...
-				MaxHeaderBytes: 512,
-				WriteTimeout:   TechnicalEndpointGeneralTimeout,
-				ReadTimeout:    TechnicalEndpointGeneralTimeout,
-			},
-			Timeout:      TechnicalEndpointGeneralTimeout,
-			TCPKeepAlive: 1 * time.Minute,
-			Logger:       graceful.DefaultLogger(),
+		srv := &http.Server{
+			Addr:           port,
+			Handler:        serveMuxHandler,
```
bieli: `port` variable is related to shared config. It will be nice to refresh `http.Server` in  `startTechnicalEndpoint`

PR #119, opened 2019-02-06
Add error column

wookie41:


storages/replicator.go
```diff
@@ -73,6 +78,29 @@ func (rc *ReplicationClient) Do(request *Request) <-chan BackendResponse {
...
+			replicatedRequest.Header.Add(headerName, headerValues[idx])
+		}
+	}
+	replicatedRequest.Host = request.Host
+	replicatedRequest.ContentLength = request.ContentLength
+	replicatedRequest.Form = request.Form
+	replicatedRequest.PostForm = request.PostForm
+	replicatedRequest.GetBody = request.GetBody
+	replicatedRequest.Trailer = request.Trailer
+	replicatedRequest.TransferEncoding = request.TransferEncoding
```
mjarco: there is shortcut for that block: 
* replicatedRequest = *request
And it looks like code duplication, see sharding/shards_ring.go#L81:112

wookie41: The shortcut would overwrite the field, including body and etc

utils/utils.go
```diff
@@ -108,4 +115,46 @@ func ExtractMultiPartUploadIDFrom(response *http.Response) (string, error) {
 		return "", errors.New("upload ID was empty")
 	}
 	return initiateMultipartUploadResult.UploadID, nil
-}
\ No newline at end of file
+}
+
+//ReplicateRequest makes a copy of the provided request
+func ReplicateRequest(ctx context.Context, request *http.Request, bodyBytes []byte) (*http.Request, error) {
```
mjarco: Move original body read here and report errors if needed

storages/replicator.go
```diff
@@ -40,18 +40,30 @@ func (rc *ReplicationClient) Do(request *Request) <-chan BackendResponse {
 	ctx, cancelFunc := context.WithCancel(newContextWithValue)
 	rc.cancelFunc = cancelFunc
 
+	bodyBytes, err := utils.ReadRequestBody(request.Request)
```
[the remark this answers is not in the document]
wookie41: Reading request's body here would require to handle to errors:
1. Error on read
2. Error on closing the original body.
The above is much cleaner for me.

utils/utils.go
wookie41: Moving the original body here would require reading the body each time we wan't to replicate the request, i.e. for each storage call.

PR #136, opened 2019-08-29
Configure ignored headers

mjarco:


internal/akubra/storages/storages.go
```diff
@@ -2,9 +2,10 @@ package storages
 
 import (
 	"fmt"
-	config2 "github.com/allegro/akubra/internal/akubra/watchdog/config"
 	"net/http"
 
+	config2 "github.com/allegro/akubra/internal/akubra/watchdog/config"
```
wookie41: watchdogConfig

PR #137, opened 2019-09-17
All shards read

wookie41:
This makes the watchdog worker delete the object versions from other shards that could have been the object's  primary shards but no longer are, for example because of a migration of weight change.


internal/brim/filter/filter_test.go
```diff
@@ -356,37 +427,44 @@ func prepareVersionMocks(bucket, key, access, secret string, fetcherMock *versio
...
 	akubraConfig := &config.Config{}
 	akubraConfig.Storages = storagesConfig.StoragesMap{}
 	akubraConfig.Shards = storagesConfig.ShardsMap{}
 
-	akubraConfig.ShardingPolicies = regionsConfig.ShardingPolicies{}
-	akubraConfig.ShardingPolicies["test"] = regionsConfig.Policies{
-		Domains: []string{"localhost"},
-		Shards:  []regionsConfig.Policy{{Weight: 1, ShardName: "test"}},
-	}
+	weightPerShard := float64(1) / float64(numberOfStoragesPerShard)
```
bieli: When value of `numberOfStoragesPerShard` will be zero result of below calculation is `+Inf`.
Maybe we need special IF condition for this case ?

ksitak: There's no test that defines it as zero.

